Issue #22, opened 2017-05-25
Like a post client


mayzabeel:
With Patch

andreldsa:
Like a post #24 
fixed posthandler merge #49

PR #49, opened 2017-05-30
fixed posthandler merge

ruanres:
Solved bug inside posthandler


andreldsa:
Like a post client #22

PR #61, opened 2017-06-02
User feedback when follow institution

andreldsa:


andreldsa:
Follow institution redesign #62

PR #97, opened 2017-06-09
User can't unfollow his intitutions

raonismaneoto:
#93


app/institution/institutionController.js
```diff
@@ -1,10 +1,12 @@
 'use strict';
 
+var institutionCtrl;
```
mayzabeel: remover

Issue #95, opened 2017-06-09
Create postDetailsDirective


mayzabeel:
used in home.html and institution_page.html

PR #103, opened 2017-06-12
Post details directive

Luiz-FS:
Issue #95


app/post/postDetailsDirective.js
```diff
@@ -0,0 +1,156 @@
+(function() {
+    'use strict';
```
ruanres: coloca fora da função

PR #125, opened 2017-06-20
Institution page redirect to home when is not new user

andreldsa:


andreldsa:
Choice of institution for new users #90

Issue #136, opened 2017-06-21
Submit comment more then once


andreldsa:
1 - Type text in the comment
2 - Press ENTER key more then once rapidly

PR #144, opened 2017-06-21
Fix submit comment more then once

andreldsa:


andreldsa:
Submit comment more then once #136

Issue #145, opened 2017-06-22
Change comment input for textarea


mayzabeel:
Change to textarea. To see big comments

Issue #142, opened 2017-06-21
Unhandled size of comment too long


andreldsa:
Write a very long comment.

`Erro {"msg": "Error! Indexed value text must be at most 1500 bytes"}`

PR #150, opened 2017-06-22
Fix size of comment and redesign comment directive

andreldsa:


andreldsa:
Unhandled size of comment too long #142

andreldsa:
Input for comments #145

PR #173, opened 2017-06-28
Test post controller

Luiz-FS:


andreldsa:
Continued on Tests for PostDirective and Post model #175

PR #266, opened 2017-08-08
Refact submit institution

mayzabeel:
When the user has some invite to create institution,  is redirect to page new invite where accept or reject the invite. 
If the invite is accepted the admin is redirect to manage institution page to edit info.
When the user accepted the invite to be admin of institution, the post method is called, so the patch method is used only to edit info of institution.

So, submitInstitution.html and submitInstitutionController.html was removed.


app/invites/newInviteController.js
```diff
@@ -13,50 +13,84 @@
...
+
         newInviteCtrl.user = AuthService.getCurrentUser();
 
         newInviteCtrl.acceptInvite = function acceptInvite(event) {
+            if (typeOfInvite === "USER") {
+                newInviteCtrl.addInstitution(event);
+            } else {
+                newInviteCtrl.updateStubInstitution();
+            }
+
```
andreldsa: rlb

app/invites/newInviteController.js
```diff
@@ -13,50 +13,84 @@
...
-                    goHome();
-                    showAlert(event, newInviteCtrl.institution.name); 
-               });
-            }, function error(response) {
-                showToast(response.data.msg);
-            });
+                promise.then(function success() {
+                    AuthService.reload().then(function() {
+                        goHome();
+                        showAlert(event, newInviteCtrl.institution.name);
```
andreldsa: messageservice

app/invites/newInviteController.js
```diff
@@ -69,21 +103,10 @@
...
-                $mdToast.simple()
-                    .textContent(msg)
-                    .action('FECHAR')
-                    .highlightAction(true)
-                    .hideDelay(5000)
-                    .position('bottom right')
-            );
-        }
-
         function showAlert(event, institutionName) {
```
andreldsa: pq tem um showAlert?

PR #270, opened 2017-08-09
Fix wrong condition to validate permission on post handler

andreldsa:
The condition to validate user permission was to have publish_post, using Utils._assert, but the that function works with the opposite condition to thrown an error. So, the correct validation is not to have publish_post.

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -105,11 +105,11 @@ def delete_all_in_index(index):
 
 class BaseHandler(webapp2.RequestHandler):
     """Base Handler."""
-    pass
-    # def handle_exception(self, exception, debug):
-    #     """Exception."""
-    #     logging.error(str(exception))
-    #     self.response.write("oops! %s\n" % str(exception))
+
+    def handle_exception(self, exception, debug):
+        """Exception."""
+        logging.error(str(exception))
+        self.response.write("oops! %s\n" % str(exception))
 
 
 class ResetHandler(BaseHandler):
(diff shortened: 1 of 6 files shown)


admin.py
```diff
@@ -105,11 +105,11 @@ def delete_all_in_index(index):
 
 class BaseHandler(webapp2.RequestHandler):
     """Base Handler."""
-
-    def handle_exception(self, exception, debug):
-        """Exception."""
-        logging.error(str(exception))
-        self.response.write("oops! %s\n" % str(exception))
+    pass
```
mayzabeel: descomentar

admin.py
```diff
@@ -340,11 +347,15 @@ def get(self):
...
         }
+
         eciis = createInstitution(data, admin)
         for user in [dalton, andre, jorge, maiana,
                      luiz, raoni, ruan, tiago, mayza, admin]:
             eciis.add_member(user)
+            user.add_institution(eciis.key)
+            user.follow(eciis.key)
 
+        print andre.permissions
```
mayzabeel: remover

admin.py
```diff
@@ -319,16 +322,20 @@ def get(self):
             'phone_number': '83 33227865',
             'state': 'active'
         }
+
         splab = createInstitution(data, admin)
         for user in [jorge, andre, admin]:
             splab.add_member(user)
+            user.add_institution(splab.key)
+            user.follow(splab.key)
+        print andre.permissions
```
mayzabeel: remover

PR #273, opened 2017-08-10
Fix unchanged list of invite after send a new one

MaianaB:


app/invites/inviteInstitutionController.js
```diff
@@ -64,6 +64,8 @@
         inviteController.sendInstInvite = function sendInstInvite(invite) {
             var promise = InviteService.sendInvite(invite);
             promise.then(function success(response) {
+                    inviteController.invite = {};
```
mayzabeel: testar

PR #324, opened 2017-08-25
Crop image in config profile controller

Luiz-FS:
I added the image cropping function on the user settings page.

### DIFF
--- a/app/auth/configProfileController.js
+++ b/app/auth/configProfileController.js
@@ -31,8 +31,8 @@
             });
         }
 
-        configProfileCtrl.cropImage = function cropImage(image_file) {
-            CropImageService.crop(image_file).then(function success(croppedImage) {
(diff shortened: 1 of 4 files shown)


app/auth/configProfileController.js
```diff
@@ -31,6 +31,14 @@
             });
         }
 
+        configProfileCtrl.cropImage = function cropImage(image_file) {
```
ruanres: imageFile

app/auth/configProfileController.js
```diff
@@ -31,6 +31,14 @@
             });
         }
 
+        configProfileCtrl.cropImage = function cropImage(imageFile) {
```
andreldsa: testes

Luiz-FS: Teste feito.

PR #352, opened 2017-09-15
Refactoring user handler

Luiz-FS:
* I removed function addInstirution on UserService.
* I removed handler put on user_handler.
* I refactored the handler patch on invite_handler.
* I Refactored the handler patch on user_handler.
* I removed chooseInstController.


handlers/invite_handler.py
```diff
@@ -34,18 +51,29 @@ def delete(self, user, key):
 
     @json_response
     @login_required
-    def patch(self, user, url_string):
+    @ndb.transactional(xg=True)
+    def patch(self, user, invite_key):
```
mayzabeel: teste

PR #354, opened 2017-09-18
Bug duplicate user using threads

ruanres:
Trying to solve the duplicate user problem using threads


andreldsa:
This PR don't solve's the problem.

PR #358, opened 2017-09-19
Refactoring model event

Luiz-FS:
* I added method __setattr__ in event model for set attributes
* I removed method treats_date of event_handler

### DIFF
--- a/models/event.py
+++ b/models/event.py
@@ -100,7 +100,12 @@ class Event(ndb.Model):
         }
 
     def __setattr__(self, attr, value):
-        """Method of set invite attributes."""
+        """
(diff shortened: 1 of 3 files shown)


models/event.py
```diff
@@ -98,3 +98,12 @@ def make(event):
             'institution_key': event.institution_key.urlsafe(),
             'key': event.key.urlsafe()
         }
+
+    def __setattr__(self, attr, value):
+        """Method of set invite attributes."""
```
mayzabeel: invite?

mayzabeel: coloca especificadamente que trata as datas do evento

PR #362, opened 2017-09-19
set admin.py to give permission analyze_request to admin

mayzabeel:


admin.py
```diff
@@ -21,6 +21,17 @@
 
 INDEX_NAME = 'institution'
 
+#  SEE HOW GET POWER USERS IDEA
```
mayzabeel: def getPowerUsers():
    userswithpermission = []
    institutionsEmpowered = Institution.query(Institution.empowered == True)
    for institution in institutionsEmpowered:
        for userKey in institution.members:
            user = userKey.get()
            if user.has_permission('analyze_request_inst',institution.key):
                userswithpermission.push(user)
    return userswithpermission

mayzabeel:
This pull request was created only to discuss about solution

PR #364, opened 2017-09-19
Login with multiple emails

andreldsa:
Loging with multiple emails allows de user to have how many emails they want and access the same data. 

The refact was pretty straightforward after realize that wasn't necessary to change the query to get a user by his email. The only thing that changed in the model it was the property email, modified to be a ndb property repeated, changing to a list, instead of a single string.

Once the model accomplished his changing, it was needed to adapt the rest of the system to interact with a user list of emails nor just a string.

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -59,7 +59,7 @@ def getGravatar(email):
     size = 40
     # construct the url
     gravatar_url = "https://www.gravatar.com/avatar/" + \
-        hashlib.md5(email.lower()).hexdigest() + "?"
+        hashlib.md5(email[0].lower()).hexdigest() + "?"
     gravatar_url += urllib.urlencode({'d': default, 's': str(size)})
     return gravatar_url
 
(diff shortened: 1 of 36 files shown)


admin.py
```diff
@@ -114,7 +114,7 @@ def delete_all_in_index(index):
 def create_profile(user, institution):
     """Create a profile."""
     profile = InstitutionProfile()
-    profile.email = user.email
+    profile.email = user.email[0]
```
mayzabeel: Tratar para quual email será enviado o request nos metodos send_email em request_institution_children.py , request_institution_parent.py a request_user.py  
Pode fazer um search por "self.admin_key.get().email" para achar mais fácil

admin.py
```diff
@@ -59,7 +59,7 @@ def getGravatar(email):
     size = 40
     # construct the url
     gravatar_url = "https://www.gravatar.com/avatar/" + \
-        hashlib.md5(email.lower()).hexdigest() + "?"
+        hashlib.md5(email[0].lower()).hexdigest() + "?"
```
mayzabeel: Essa lista de emails do usuário poderia ser atualizada quando ele preenche o perfil institutucional e insere um novo email né?

admin.py
andreldsa: resolvido.

admin.py
andreldsa: Manter o email que está no índice 0 da lista de emails, que é o principal.

PR #380, opened 2017-09-25
fixing bug in edit profile when the institution is new

tiagolimpe:
Before this change, edit profile not work when the institution is new. Now user can edit profile in this case.


app/invites/newInviteController.js
```diff
@@ -75,6 +75,7 @@
                     newInviteCtrl.user.institutions.push(institutionSaved);
                     newInviteCtrl.user.institutions_admin.push(institutionSaved.key);
                     newInviteCtrl.user.follow(institutionSaved);
+                    newInviteCtrl.user.addProfile(createNewProfile(institutionSaved));
```
mayzabeel: teste

PR #394, opened 2017-09-27
Fixing the self notification bug in shared-posts

tiagolimpe:
<p><b>Feature/Bug description:</b> When user is author of post and shared the post, the application send notification to own user.</p>

<p><b>Solution:</b> Added verification if user.key is equal shared_post.author in post_collection_handler in POST request.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/handlers/post_collection_handler.py
+++ b/handlers/post_collection_handler.py
@@ -63,7 +63,7 @@ class PostCollectionHandler(BaseHandler):
                         entity_type,
                         post.key.urlsafe())
 
-            if(post.shared_post):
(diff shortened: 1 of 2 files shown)


handlers/post_collection_handler.py
```diff
@@ -63,7 +63,7 @@ def post(self, user):
                         entity_type,
                         post.key.urlsafe())
 
-            if(post.shared_post):
+            if(post.shared_post and user.key != post.shared_post.get().author):
```
andreldsa: faz teste, pra garantir que n vai mais acontecer

PR #397, opened 2017-09-28
Fix institutions rejected invites

ruanres:
<p><b>Bug description:</b> Rejected institution invites were being shown as pending invites when the user tries to create an institution invite with the same name as one that has already been rejected.</p>

<p><b>Solution:</b> On invite handler delete method, when an invite is deleted, the corresponding stub institution has its state changed to inactive as well as its document.</p>

### DIFF
--- a/app/invites/suggestInstitutionController.js
+++ b/app/invites/suggestInstitutionController.js
@@ -13,6 +13,7 @@
         suggestInstCtrl.requested_invites = requested_invites;
         suggestInstCtrl.isHierarchy = isHierarchy;
         var ACTIVE_STATE = "active";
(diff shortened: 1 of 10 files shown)


app/invites/suggestInstitutionController.js
```diff
@@ -22,15 +23,19 @@
             $mdDialog.cancel();
         };
 
-        suggestInstCtrl.showSuggestion = function(state) {
+        suggestInstCtrl.showSuggestion = function showSuggestion(state) {
             var isActive = state === ACTIVE_STATE;
```
andreldsa: usa a função abaixo, isActive.

PR #400, opened 2017-10-02
Sending notification when invite or request is accepted or rejected

tiagolimpe:
<p><b>Feature/Bug description:</b> When invites or requests is accepted or rejected the sender is not notified when is accepted or rejected.</p>

<p><b>Solution:</b> When user or admin accept or reject invite (or request), invite_handler send notification to sender of invite.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/handlers/institution_children_request_handler.py
+++ b/handlers/institution_children_request_handler.py
@@ -33,6 +33,8 @@ class InstitutionChildrenRequestHandler(BaseHandler):
         institution_children.parent_institution = request.institution_key
         institution_children.put()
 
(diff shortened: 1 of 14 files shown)


handlers/institution_children_request_handler.py
```diff
@@ -33,6 +33,8 @@ def put(self, user, request_key):
         institution_children.parent_institution = request.institution_key
         institution_children.put()
 
+        request.send_response_notification(user, request.admin_key.urlsafe(), 'ACCEPT_INSTITUTION_LINK')
```
andreldsa: massa, faz um teste para cada um dessas funções que tu modificou, verificando se o send_notification foi chamada.

PR #402, opened 2017-10-02
Confirm authorization user for create institution

MaianaB:
<p><b>Feature/Bug description:</b> Before create institution the user should confirm that can create.</p>

<p><b>Solution:</b> Added checkbox that user affirms that can create institution. </p>

### DIFF
--- a/app/styles/custom.css
+++ b/app/styles/custom.css
@@ -357,6 +357,5 @@ a:active {
 
 .green-button {
     background-color: #8BC34A;
-    color: #FFFFFF
-;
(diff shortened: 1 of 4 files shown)


app/styles/custom.css
```diff
@@ -353,4 +353,10 @@ a:active {
 
 .inst-timeline-center {
     margin-left: 46%;
+}
+
+.green-button {
+    background-color: #8BC34A;
+    color: #FFFFFF
+;
```
andreldsa: ponto e vírgula no canto errado.

PR #405, opened 2017-10-03
sending email to notify if user is accepted or rejected to be member of institution

tiagolimpe:
<p><b>Feature/Bug description:</b> When new user in platform send request to be member of existing institution, the platform not notify if user is accepted or rejected.</p>

<p><b>Solution:</b> Now the platform send email to notify when user is accepted or rejected to be member of requested institution.</p>

<p><b>TODO/FIXME:</b> n/a</p>


backend/service_messages.py
```diff
@@ -1,6 +1,11 @@
+# -*- coding: utf-8 -*-
 """Service of send messages."""
+import sys
 from google.appengine.api import taskqueue
 
+reload(sys)
```
andreldsa: colocar TODO p remover isso quando os textos em português dos emails forem extraídos do código.

PR #418, opened 2017-10-05
fix bug user with unknown name in notification

tiagolimpe:
<p><b>Feature/Bug description:</b> When user is invited to be admin of an institution, if the user accept or reject the invite, the notification sent to sender of invite is displayed with Unknown name.</p>

<p><b>Solution:</b> In case of reject invite, the user state is inactive therefore the name displayed in the notification is the user email. In case of acceptance, the correct name is displayed.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/invites/newInviteController.js
+++ b/frontend/invites/newInviteController.js
@@ -143,14 +143,15 @@
         function deleteInvite() {
             var promise = InviteService.deleteInvite(newInviteCtrl.inviteKey);
             promise.then(function success() {
-                AuthService.reload().then(function() {
-                    if(newInviteCtrl.user.isInactive()) {
-                        $state.go("user_inactive");
-                    } else {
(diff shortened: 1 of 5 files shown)


frontend/invites/newInviteController.js
```diff
@@ -143,14 +143,15 @@
         function deleteInvite() {
             var promise = InviteService.deleteInvite(newInviteCtrl.inviteKey);
             promise.then(function success() {
-                AuthService.reload().then(function() {
-                    if(newInviteCtrl.user.isInactive()) {
-                        $state.go("user_inactive");
-                    } else {
+                if (newInviteCtrl.isInviteUser()) {
+                    AuthService.reload().then(function() {
```
andreldsa: tem certeza q precisa do reload? não pode ser um save não?

PR #427, opened 2017-10-16
Fixing inviteIsMember and inviteeIsInvited functions

raonismaneoto:
When the admin tried to invite an user that had already been invited or that was a member he was redirected to the error page because the frontend wasn't catching the error.
</p>

<p><b>Solution:</b>
_.includes and _.some weren't doing what they were supposed to do when were implemented. Then, I've fixed inviteeIsMember and inviteeIsInvited, the functions that used to use the first ones.
</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/institution/managementMembersController.js
+++ b/frontend/institution/managementMembersController.js
@@ -3,7 +3,7 @@
     var app = angular.module('app');
 
     app.controller("ManagementMembersController", function InviteUserController(
-        InviteService, $mdToast, $state, $mdDialog, InstitutionService, AuthService, MessageService, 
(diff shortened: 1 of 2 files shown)


frontend/institution/managementMembersController.js
```diff
@@ -130,11 +130,15 @@
         }
 
         function inviteeIsMember(invite) {
-            return _.includes(_.map(manageMemberCtrl.members, getEmail), invite.invitee);
+            return _.find(_.map(manageMemberCtrl.members, getEmail), function(emails) {
```
andreldsa: testes

PR #462, opened 2017-10-26
adding specific expected exception in post_handler_test

tiagolimpe:
<p><b>Feature/Bug description:</b> When patch_method in post_handler_test throw an error, this error message is not verified.</p>

<p><b>Solution:</b> Verifying if the specific message is thrown when the method crash.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/backend/test/post_handler_test.py
+++ b/backend/test/post_handler_test.py
@@ -85,18 +85,18 @@ class PostHandlerTest(TestBaseHandler):
         expected_alert = "Expected: " + exception_message + ". But got: "
 
         # Call the patch method and assert that  it raises an exception
-        with self.assertRaises(Exception) as ex:
+        with self.assertRaises(Exception) as raises_context:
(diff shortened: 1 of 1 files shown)


backend/test/post_handler_test.py
```diff
@@ -80,13 +80,24 @@ def test_delete(self, verify_token):
     @patch('utils.verify_token', return_value={'email': 'first_user@gmail.com'})
     def test_patch(self, verify_token):
         """Test the post_handler's patch method."""
+
+        exception_message = "Error! User is not allowed to edit this post"
+        expected_alert = "Expected: " + exception_message + ". But got: "
+
         # Call the patch method and assert that  it raises an exception
-        with self.assertRaises(Exception):
+        with self.assertRaises(Exception) as ex:
```
andreldsa: renomeia ex para raises_context

PR #485, opened 2017-11-06
Verify user email with firebase

andreldsa:
<p><b>Feature/Bug description:</b> Email verification with firebase. Once the user creates it's account an email is sent to him with a link to verify his account.</p>

<p><b>Solution:</b> Used the firebase native function to send the verification email.</p>

<p><b>TODO/FIXME:</b> This verification breaks the test fake login.</p>

### DIFF
--- a/frontend/auth/authService.js
+++ b/frontend/auth/authService.js
@@ -1,7 +1,5 @@
 'use strict';
 
-var out_user;
-
 (function() {
     var app = angular.module("app");
 
--- a/frontend/auth/authService.js
+++ b/frontend/auth/authService.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var out_user;
+
 (function() {
     var app = angular.module("app");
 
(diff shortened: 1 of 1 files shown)


frontend/auth/authService.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+var out_user;
```
mayzabeel: não tá sendo usada

frontend/auth/authService.js
```diff
@@ -131,6 +139,15 @@
             return deferred.promise;
         };
 
+        service.sendEmailVerification = function sendEmailVerification() {
+            authObj.$getAuth().sendEmailVerification().then(
+            function success() {
+                MessageService.showToast('Email de verificação enviado para o seu email.');
+            }, function error(error) {
+                console.error(error);
```
mayzabeel: remover

andreldsa: Esse console é de erro, caso algo dê errado com o envio do email.

PR #495, opened 2017-11-08
New layout of login page and registration page

Luiz-FS:
<p><b>Feature/Bug description:</b> Create a new layout for the login page and the registration page.</p>

<p><b>Solution:</b> Using the angular directives I created the new layout according to what was passed by the design.</p>

<p><b>TODO/FIXME:</b> n/a</p>


frontend/styles/custom.css
```diff
@@ -322,7 +322,7 @@ a:active {
 }
 
 .youtube-player img:hover {
-    
+
```
andreldsa: rlb

PR #504, opened 2017-11-10
Bug on showing the name of who deleted the post

ruanres:
<p><b>Bug description:</b> When the post was deleted, the last_modified_by property was not been updated on frontend. So, it was been showing the name of the last person who modified the post and not the user who deleted it</p>

<p><b>Solution:</b> Set post property last_modified_by to current user name after the post is successfully  deleted</p>

### DIFF
--- a/frontend/test/specs/postSpec.js
+++ b/frontend/test/specs/postSpec.js
@@ -125,4 +125,18 @@ describe('Test Post Model:', function() {
             expect(post.hasImage()).toBe(false);
         });
     });
+
+    describe('remove()', function() {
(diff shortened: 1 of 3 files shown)


frontend/test/specs/postSpec.js
```diff
@@ -125,4 +125,18 @@ describe('Test Post Model:', function() {
             expect(post.hasImage()).toBe(false);
         });
     });
+
+    describe('remove()', function() {
+        it('should set the post state to false', function() {
```
mayzabeel: state to deleted

PR #513, opened 2017-11-16
Fix user request

tiagolimpe:
<p><b>Feature/Bug description:</b> When the user request to be member of institution more than once the message "500 : Error! The sender is already invited" is displayed and the system stop.</p>

<p><b>Solution:</b> Added verification if the emails of user is contained in invitations with sent status for institution.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/requests/requestInvitationController.js
+++ b/frontend/requests/requestInvitationController.js
@@ -14,6 +14,7 @@
         requestInvCtrl.wasSearched = false;
         requestInvCtrl.canCreate = false;
         requestInvCtrl.currentUser = AuthService.getCurrentUser();
+        requestInvCtrl.requestsOfSelectedInst = [];
         var ACTIVE = 'active';
 
         requestInvCtrl.sendRequest = function sendRequest() {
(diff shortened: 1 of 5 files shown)


frontend/requests/requestInvitationController.js
```diff
@@ -38,6 +39,26 @@
             });
         };
 
+        requestInvCtrl.verifyAndSendRequest = function verifyAndSendRequest() {
```
mayzabeel: teste

frontend/test/specs/requestInvitationControllerSpec.js
```diff
@@ -103,6 +103,27 @@
             });
         });
 
+        describe('verifyAndSendRequest()', function() {
+            beforeEach(function() {
+                requestInvCtrl.requestsOfSelectedInst = [{sender_key: maiana.key, status: 'sent'}];
```
andreldsa: utiliza variáveis com nomes genéricos, user, other_user, etc.

PR #515, opened 2017-11-17
Fix bug add media when edit post and it has media.

MaianaB:
<p><b>Feature/Bug description:</b> When edit post and it has image or video, the user can add other media.</p>

<p><b>Solution:</b> Added verification if has image and set value of video when call method createEditedPost.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/post/postDirective.js
+++ b/frontend/post/postDirective.js
@@ -26,7 +26,7 @@
         var observer;
 
         postCtrl.hasMedia = function hasMedia() {
-            return postCtrl.photoBase64Data || postCtrl.pdfFiles.length > 0 || postCtrl.addVideo;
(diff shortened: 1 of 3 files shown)


frontend/post/postDirective.js
```diff
@@ -69,6 +69,7 @@
         postCtrl.createEditedPost = function createEditedPost(post) {
             postCtrl.photoUrl = post.photo_url;
             postCtrl.pdfFiles = post.pdf_files.slice();
+            postCtrl.addVideo = post.video_url ? true : false;
```
mayzabeel: hasVideo não é melhor não?

andreldsa: é

PR #521, opened 2017-11-20
fixing showing files when post is changed of common to survey

tiagolimpe:
<p><b>Feature/Bug description:</b> When the type of post is changed of common to survey, the files in post keep displayed in post.</p>

<p><b>Solution:</b> Added different behavior in clearPost function when type of post is common or survey.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/post/postDirective.js
+++ b/frontend/post/postDirective.js
@@ -239,7 +239,7 @@
         };
 
         postCtrl.clearPost = function clearPost() {
-            postCtrl.typePost === "Common" ? postCtrl.post = {}: ()=>{};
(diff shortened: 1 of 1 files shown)


frontend/post/postDirective.js
```diff
@@ -239,11 +239,10 @@
         };
 
         postCtrl.clearPost = function clearPost() {
-            postCtrl.post = {};
+            postCtrl.typePost == "Common" ? postCtrl.post = {}: ()=>{};
```
mayzabeel: ===

PR #523, opened 2017-11-20
bug on showing institution requests

ruanres:
<p><b>Bug description:</b> The institution requests was not been shown because the response.data is undefined </p>

<p><b>Solution:</b> The response from getRequestsInst does not have data, so, it was removed and the requests are been get directly from the response. </p>

### DIFF
--- a/frontend/invites/inviteInstitutionController.js
+++ b/frontend/invites/inviteInstitutionController.js
@@ -111,7 +111,7 @@
 
         function loadSentRequests() {
             RequestInvitationService.getRequestsInst().then(function success(response) {
-                inviteController.sent_requests = response.data;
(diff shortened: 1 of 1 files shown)


frontend/invites/inviteInstitutionController.js
```diff
@@ -111,7 +111,7 @@
 
         function loadSentRequests() {
             RequestInvitationService.getRequestsInst().then(function success(response) {
-                inviteController.sent_requests = response.data;
+                inviteController.sent_requests = response;
```
mayzabeel: resolver no servidor

andreldsa: muda de response para o nome da instância do objeto.

PR #525, opened 2017-11-20
fixing delete event bug

tiagolimpe:
<p><b>Feature/Bug description:</b> When event is deleted his not disappear on events timeline.</p>

<p><b>Solution:</b> Added a filter in array of events to remove deleted events when the delete function is called.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/backend/handlers/event_handler.py
+++ b/backend/handlers/event_handler.py
@@ -43,6 +43,7 @@ class EventHandler(BaseHandler):
             event_json
         ))
 
+    @json_response
     @login_required
     @is_event_author
     def delete(self, user, key):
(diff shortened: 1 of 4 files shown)


backend/handlers/event_handler.py
```diff
@@ -43,6 +43,7 @@ def get(self, user, url_string):
             event_json
         ))
 
+    @json_response
```
mayzabeel: precisa?

frontend/event/eventController.js
```diff
@@ -108,8 +112,10 @@
         };
 
         eventCtrl.isLongText = function isLongText(event){
-            var numberOfChar = event.text.length;
-            return numberOfChar >= LIMIT_CHARACTERS;
+            if(event.text) {
```
mayzabeel: todo evento não tem text não?

PR #527, opened 2017-11-20
Create user template route inheritance

andreldsa:
<p><b>Feature/Bug description:</b> Some pages of the system, like config profile and invite institutions,
 were implemented without the left navbar of the user.</p>

<p><b>Solution:</b> Create a new abstract route (app.user), that encapsulate the left navbar and pass the user_content route view. That way, the routes that needed the left navbar just inherit from app.user and implement the user_content HTML.</p>

<p><b>TODO/FIXME:</b> n/a</p>


frontend/invites/invite_institution.html
```diff
@@ -1,6 +1,6 @@
-<md-content layout="row" md-colors="{background: 'grey-200'}" layout-align="center" class="body">
-  <div flex-gt-lg="80" flex-lg="90" layout="row" layout-align="center" layout-fill>
-    <div flex="80">
+<md-content layout="row" layout-align="center" class="body hide-scrollbar">
```
mayzabeel: precisa da cor de background aqui não?

andreldsa: Já tem o background do layout pai (app.user)

PR #538, opened 2017-11-21
 Fix User Link Removal

Luiz-FS:
<p><b>Feature/Bug description:</b> Every time a user removes a connection to a facility, a logout is made.</p>

<p><b>Solution:</b> Check if the user has more than one institution, if so, only removes the link, otherwise logout.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/auth/configProfileController.js
+++ b/frontend/auth/configProfileController.js
@@ -153,7 +153,6 @@
                 _.remove(configProfileCtrl.newUser.institution_profiles, function(profile) {
                     return profile.institution_key === institution_key;
                 });
-
                 AuthService.save();
             } else {
                 AuthService.logout();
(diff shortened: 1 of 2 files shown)


frontend/auth/configProfileController.js
```diff
@@ -138,13 +138,28 @@
...
 
+        function removeConection(institution_key) {
+            if (configProfileCtrl.newUser.institutions.length > 1) {
+                _.remove(configProfileCtrl.newUser.institutions, function(institution) {
+                    return institution.key === institution_key;
+                });
+                _.remove(configProfileCtrl.newUser.institution_profiles, function(profile) {
+                    return profile.institution_key === institution_key;
+                });
+
```
mayzabeel: rlb

frontend/auth/configProfileController.js
```diff
@@ -138,13 +138,28 @@
...
             promise.then(function success() {
-                AuthService.logout();
+                removeConection(institution_key);
             }, function error(response) {
                 MessageService.showToast(response.data.msg);
             });
             return promise;
         }
 
+        function removeConection(institution_key) {
```
andreldsa: tests

PR #550, opened 2017-11-24
Fix remove child institution

raonismaneoto:
The parent admin couldn't remove his child institution because it was missing a verification in is_admin decorator. Besides, removeProfile hasn't been called when a child institution was removed.
</p>

<p><b>Solution:</b>
I've added a verification that checks if the user is parent admin of the institution he wants to remove, and called removeProfile in removeChildInst.
</p>

<p><b>TODO/FIXME:</b> n/a</p>


backend/handlers/institution_handler.py
```diff
@@ -56,8 +56,10 @@ def check_authorization(self, user, institution_key, *args):
 
         userisNotAdminOfInstitution = institution.key not in user.institutions_admin
         institutionisNotManagedByUser = institution.admin != user.key
+        userIsNotParentAdmin = institution.parent_institution not in user.institutions_admin
 
-        Utils._assert(userisNotAdminOfInstitution or institutionisNotManagedByUser,
+        Utils._assert(userIsNotParentAdmin and
```
mayzabeel: teste

PR #574, opened 2017-11-30
Changing display of institution suggestions

tiagolimpe:
<p><b>Feature/Bug description:</b> When the admin of institution invite institution without hierarchy the dialog displayed is the same when the invite is to hierarchy. Also, button "ver detalhes" is behaving unexpectedly, not redirecting to institution page.</p>

<p><b>Solution:</b> Changing the display to be different in case of invite are without hierarchy and delegate the responsibility of redirect to institution page to $state of ui-router library.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/invites/suggestInstitutionController.js
+++ b/frontend/invites/suggestInstitutionController.js
@@ -16,7 +16,8 @@
         var PENDING_STATE = "pending";
 
         suggestInstCtrl.goToInstitution = function goToInstitution(institutionKey) {
-            window.open(makeUrl(institutionKey), '_blank');
+            const url = $state.href('app.institution', {institutionKey: institutionKey});
+            window.open(url, '_blank');
         };
 
         suggestInstCtrl.cancel = function cancel() {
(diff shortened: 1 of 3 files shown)


frontend/invites/suggestInstitutionController.js
```diff
@@ -51,14 +51,22 @@
...
-            } else {
-                message = 'A instituição que você quer convidar é alguma dessas?';
-            }
-            return message;
+        suggestInstCtrl.titleMessage = function() {
+            const oneInst = 'Já existe uma instituição com esse nome.';
+            const moreThanOneInst = 'Já existem instituições com esse nome.';
+            const question = ' Deseja convidar mesmo assim?';
+            const selectMsg = 'instituição existente se deseja convidá-la ou selecione a opção de convidar uma nova instituição.'
+            return suggestInstCtrl.institutions.length === 1 ?
```
andreldsa: está bem confuso essa lógica com operador ternário e concatenação de strings.

PR #577, opened 2017-11-30
Implement infrastructure based on Jenkins pipeline with Docker containers

andreldsa:
<b>Feature/Bug description:</b> In the previous infrastructure architecture we were using the Travis CI to run the tests in the cloud during the development process, however, we had a lot of disconnected tests and inconsistencies between the cloud and the local environments, causing a high cost to maintain those pipelines.

<b>Solution:</b> Implement a new infrastructure based on Jenkins pipeline with Docker containers to simulate the local environment in the cloud. This PR consists of the files need to the cloud infrastructure to up and run the test pipeline. The image used it's here: https://hub.docker.com/r/andreldsa/nodejs-python/

It was created a new script, `setup_env_test`, that is responsible for setup the backend environment to run the **Pyunit** tests, and it's used only by the **ecis script**. The backend test now is running inside a **virtualenv**, that is a closed "container" that has all the capabilities to execute python programs. Using virtualenv we separate the system/local packages from the packages used by the platform, avoiding root packages installations and inconsistencies between machines and cloud infra.

We have a **new feature** in the ecis script, now we can type **./ecis test server --clean** to clean up the backend environment. This feature is **used by the Jenkins**, to start a new setup of test and avoid packages errors. In the **development process**, this could be useful when the backend tests display some **packages errors**.

This new architecture needed a creation of a new Github repository, [Google App Engine](https://github.com/eciis/google_appengine), to provide our script's to pull the packages from a reliable source without HTTP downloads. Also, it was created a script inside that repository to eventually update the SDK and push the changes to it.

ecis-infra


Jenkinsfile
```diff
@@ -0,0 +1,44 @@
...
+  stages {
+    stage('Tests') {
+      steps {
+        parallel(
+          "Backend": {
+            sh './ecis test server --clean'
+            
+          },
+          "Frontend": {
+            sh '''
```
raonismaneoto: How will we run the client's tests now?

andreldsa: With the command `./ecis test client` in the src of the project.

PR #581, opened 2017-12-01
Verify permissions and throw exception in method has_permission of user

mayzabeel:
<p><b>Feature/Bug description:</b>Check permissions with denying the condition </p>

<p><b>Solution:</b> Refact method has permission of user</p>

<p><b>TODO/FIXME:</b> Chenge verifications of super users to new mode</p>

### DIFF
--- a/backend/handlers/institution_hierarchy_handler.py
+++ b/backend/handlers/institution_hierarchy_handler.py
@@ -24,11 +24,11 @@ class InstitutionHierarchyHandler(BaseHandler):
     @login_required
     @ndb.transactional(xg=True)
     def delete(self, user, institution_key, institution_link):
+        """Handle delete link between institutions."""
         user.has_permission('remove_link',
                             "User is not allowed to remove link between institutions",
                             institution_key)
 
-        """Handle delete link between institutions."""
         is_parent = self.request.get('isParent')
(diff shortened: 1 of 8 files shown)


backend/test/institution_member_handler_test.py
```diff
@@ -115,11 +115,15 @@ def test_delete_not_admin(self, verify_token):
         self.assertTrue(self.institution.key in self.second_user.institutions,
                         "Institution should be in institutions of second_user")
         # Call the delete method
+        print "MAYZAAAAAA 4444444444"
```
andreldsa: remove

backend/handlers/institution_hierarchy_handler.py
```diff
@@ -24,9 +24,9 @@ class InstitutionHierarchyHandler(BaseHandler):
     @login_required
     @ndb.transactional(xg=True)
     def delete(self, user, institution_key, institution_link):
-        Utils._assert(not user.has_permission('remove_link', institution_key),
-                      "User is not allowed to remove link between institutions",
-                      NotAuthorizedException)
+        user.has_permission('remove_link',
```
andreldsa: This line must be after the function comment (line 31).

PR #584, opened 2017-12-04
Refact UserProfile is_valid method

ruanres:
<p><b>Feature/Bug description:</b> Now the is_valid method only checks the last profile added</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/backend/models/user.py
+++ b/backend/models/user.py
@@ -61,8 +61,9 @@ class InstitutionProfile(ndb.Model):
     def is_valid(profiles):
         """Verify the new user profile."""
         new_profile = profiles[len(profiles) - 1]
-        office_is_defined = new_profile.office != None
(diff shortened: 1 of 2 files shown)


backend/models/user.py
```diff
@@ -59,11 +59,10 @@ def make(self):
...
     @staticmethod
     def is_valid(profiles):
-        """Verify the user profile."""
-        for profile in profiles:
-            if not profile.office:
-                return False
-        return True
+        """Verify the new user profile."""
+        new_profile = profiles[len(profiles) - 1]
+        office_is_defined = new_profile.office != None
```
andreldsa: Instead of compare with None, use if statement:

```
if (new_profile.office):
    return true
return false
```

PR #586, opened 2017-12-05
Permission edit delete post

MaianaB:
<p><b>Feature/Bug description:</b> The method is_authorized verify if user be author and admin of institution.</p>

<p><b>Solution:</b> Added permissions and methods that verify if user has permission.</p>

<p><b>TODO/FIXME:</b> n/a</p>


backend/utils.py
```diff
@@ -222,7 +205,7 @@ def has_permission(permission_type):
     """Check user permission."""
     def method_for_verification(method):
         def check_permission(self, user, *args):
-            Utils._assert( (permission_type not in user.permissions),
+            Utils._assert((permission_type not in user.permissions),
```
andreldsa: Use user.check_permission

PR #603, opened 2017-12-12
Refact to check permissions instead of use decorators

mayzabeel:
<p><b>Feature/Bug description:</b>Refact to check permissions instead of use decorators is_admin and is_admin_of_requested_inst</p>

<p><b>Solution:</b>Create permissions to user admin: <br>
 "send_link_inst_request" --> the user can send requests for links with other institutions <br>
"send_link_inst_invite" --> the user can send invite to new institutions that are connect with your institution <br>
"invite_members" --> the user can invite members to your institution <br>
"answer_user_request" --> the user can aprove or not requests of users to be members of institution <br>
</p>

<p><b>TODO/FIXME:</b> n/a</p>


backend/test/request_handler_test.py
```diff
@@ -26,56 +30,90 @@ def setUp(cls):
...
         cls.testapp = cls.webtest.TestApp(app)
-        initModels(cls)
 
-    @patch('utils.verify_token', return_value={'email': 'useradmin@test.com'})
+    @patch('utils.verify_token', return_value=ADMIN)
     def test_get(self, verify_token):
         """Test method get of RequestHandler."""
-        response = self.testapp.get('/api/requests/' + self.request.key.urlsafe() + '/user')
-        request = json.loads(response._app_iter[0])
+        admin = mocks.create_user(ADMIN['email'])
```
andreldsa: We need to think how to solve the recurrent repetition of code even after mocks creation. I think we could create some scenarios like an institution with admin and hierarchy.

PR #617, opened 2017-12-19
Fix bug to reply comment

MaianaB:
<p><b>Feature/Bug description:</b> Bug in reply comment, the condition in html that show the div was wrong.</p>

<p><b>Solution:</b> The condition was "showReply && !commentCtrl.canReply() && !commentCtrl.isInstInactive()" it changed to "commentCtrl.canReply()". The method canReply() in controller was modified to verify all conditions needed that user can reply.</p>

<p><b>TODO/FIXME:</b> </p>

### DIFF
--- a/frontend/post/postDetailsDirective.js
+++ b/frontend/post/postDetailsDirective.js
@@ -619,7 +619,8 @@
         };
 
         commentCtrl.canReply = function canReply() {
-            return commentCtrl.showReplies && !commentCtrl.isDeletedPost() && !commentCtrl.isInstInactive();
+            return commentCtrl.showReplies && !commentCtrl.isDeletedPost() &&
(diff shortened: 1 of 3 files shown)


frontend/post/postDetailsDirective.js
```diff
@@ -615,20 +619,19 @@
         };
 
         commentCtrl.canReply = function canReply() {
-            return commentCtrl.post.state === 'deleted';
+            return commentCtrl.showReplies && !commentCtrl.isDeletedPost() && !commentCtrl.isInstInactive();
```
andreldsa: This change should break some tests. If none is broken or we don't have tests to cover that function or those that exist are wrong.

MaianaB: there was no test for the method

PR #620, opened 2017-12-20
added currentInstitutionKey as a like request params

ruanres:
<p><b>Feature/Bug description:</b>With the new notification design, it is necessary to show the user who liked the post as well as the institution name he was logged in at the moment. In the previous implementation of the like post action, in both client and server, that was not possible. </p>

<p><b>Solution:</b>On client, the method likePost sends the current institution name inside a json to the server. So, the LikeHandler retrieves that object and forwards it to the service_message module to create the like notification with the institution name</p>

<p><b>TODO:</b> Use this solution on another handlers that send notifications</p>


backend/handlers/like_handler.py
```diff
@@ -46,9 +46,11 @@ def post(self, user, post_key, comment_id=None, reply_id=None):
         """This method is only meant to give like in post."""
         post = ndb.Key(urlsafe=post_key).get()
         institution = post.institution.get()
+        current_institution_key = self.request.get('currentInstKey')
```
andreldsa: Why not just pass the institution name? This approach it's going to be expensive, once we have to get the institution from the database then get the name. A good idea could be, pass a JSON institution just with the name, because, if we need to extend the data passed through the request it's not necessary to change the data structure.

backend/service_messages.py
```diff
@@ -11,14 +11,16 @@
 reload(sys)
 sys.setdefaultencoding('utf8')
 
-def create_message(sender_key):
+def create_message(sender_key, current_institution_key):
```
andreldsa: docstring

frontend/post/postDetailsDirective.js
```diff
@@ -226,7 +226,8 @@
 
         function likePost() {
             postDetailsCtrl.savingLike = true;
-            var promise = PostService.likePost(postDetailsCtrl.post);
+            var currentInstitutionKey = postDetailsCtrl.user.current_institution.key;
+            var promise = PostService.likePost(postDetailsCtrl.post, currentInstitutionKey);
```
andreldsa: Tests

frontend/post/postService.js
```diff
@@ -42,9 +42,10 @@
             return deferred.promise;
         };
 
-        service.likePost = function likePost(post) {
+        service.likePost = function likePost(post, currentInstitutionKey) {
```
andreldsa: Tests

backend/service_messages.py
```diff
@@ -42,7 +44,7 @@ def create_entity(entity_key):
     return json.dumps(entity)
 
 
-def send_message_notification(receiver_key, sender_key, entity_type, entity_key):
+def send_message_notification(receiver_key, sender_key, entity_type, entity_key, current_institution_key=None):
```
andreldsa: tests

backend/service_messages.py
andreldsa: tests

backend/handlers/like_handler.py
```diff
@@ -46,7 +46,7 @@ def post(self, user, post_key, comment_id=None, reply_id=None):
         """This method is only meant to give like in post."""
         post = ndb.Key(urlsafe=post_key).get()
         institution = post.institution.get()
-        current_institution_key = self.request.get('currentInstKey')
+        current_institution = self.request.body
```
andreldsa: This line doesn't make sense, the body is the structure that encapsulates the date coming from the request. If you're passing the current institution in the body you should create a property current_institution inside the body structure and assign the date to it.

PR #628, opened 2017-12-21
View members of institution

MaianaB:
<p><b>Feature/Bug description:</b> Create a view member institution, that is visible to all user of the platform. The page contains first the admin of institution then all members of the institution, the user can search by name. </p>

<p><b>Solution:</b> Create using design for all size.</p>

member

corpo

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/app.js
+++ b/frontend/app.js
@@ -112,6 +112,15 @@
                     }
                 }
             })
+            .state("app.institution.members", {
+                url: "/institution/:institutionKey/members",
+                views: {
+                    institution_content: {
+                        templateUrl: "app/institution/members.html",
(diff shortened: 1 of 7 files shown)


frontend/app.js
```diff
@@ -112,6 +112,15 @@
                     }
                 }
             })
+            .state("app.institution.members", {
+                url: "/institution/:institutionKey/members",
+                views: {
+                    institution_content: {
+                        templateUrl: "app/institution/members.html",
+                        controller: "ManagementMembersController as membersCtrl"
```
mayzabeel: In institutionController remove method getMembers because is unnecessary

PR #639, opened 2018-01-08
Send Institution name in like comments requests

ruanres:
<p><b>Feature/Bug description:</b> When the user likes a comment, the name of his current Institution is not being send in the like request.</p>

<p><b>Solution:</b> It was modified the like method, from CommentService, to send the user current institution name in the request</p>

<p><b>TODO/FIXME:</b> n/a</p>


backend/test/like_comment_handler_test.py
```diff
@@ -0,0 +1,152 @@
...
+            len(likes), 0,
+            "The number of likes should be zero, but it was %s" % len(likes)
+        )
+
+    def tearDown(cls):
+        """Deactivate the test."""
+        cls.test.deactivate()
+
+
+def init(cls):
```
andreldsa: Remove init() and move the creation of mock to inside each test, only with the required entities.

backend/test/like_post_handler_test.py
```diff
@@ -138,26 +154,10 @@ def tearDown(cls):
 def initModels(cls):
```
andreldsa: Same.

PR #647, opened 2018-01-10
Accept Invite page

andreldsa:
<p><b>Feature/Bug description:</b> Implement a page on the frontend to display an invite, when the user access directly from the email or URL.</p>

<p><b>Solution:</b> Added a new router as "accept_invite" that receives a query parameter called id.
 The id refers to the key of the invite, and it's used to retrieve the data from the server to load the page.
 If the id point to an invite that is already responded, the state is redirected to "signin state".

The accept invite page has two buttons, "not user" and "log in". If the user hasn't an account it's redirected to a signup page, with the email already filled, to prevent change the email from the invite.

screenshot from 2018-01-15 18-03-25

screenshot from 2018-01-15 18-03-51

screenshot from 2018-01-15 18-05-13

</p>

<p><b>TODO/FIXME:</b> n/a</p>


backend/handlers/invite_handler.py
```diff
@@ -7,6 +7,7 @@
 from utils import login_required
 from handlers.base_handler import BaseHandler
 from models.user import InstitutionProfile
+from models.invite import Invite
```
mayzabeel: Massa, pelo que vi falta só colocar o link no email com o convite né?

PR #657, opened 2018-01-15
Get institution's events in the server

raonismaneoto:
Get institution's events in the server.
</p>

<p><b>Solution:</b>
I've created a handler that handles institution's events get requests.
</p>

<p><b>TODO/FIXME:</b> n/a</p>


backend/utils.py
```diff
@@ -174,6 +175,25 @@ def offset_pagination(page, number_fetchs, query):
     return [query, more]
 
 
+def paginate(params, query):
```
andreldsa: A better name is query_paginated. Add a well-explained docstring, renaming params to something more explicit and explain in the docstring what it receives. This function should receive page and limit of elements as parameters as optional.

backend/utils.py
```diff
@@ -175,18 +175,28 @@ def offset_pagination(page, number_fetchs, query):
     return [query, more]
 
 
-def paginate(params, query):
-    """Paginate events."""
-    params = dict(params)
+def query_paginated(request_data, query, page=None, limit=None):
```
andreldsa: page=Utils.DEFAULT_PAGINATION_OFFSET

That already assign the default value to the variable if None is coming from the call.

PR #662, opened 2018-01-16
Add current institution name to comments and replies requests

ruanres:
<p><b>Feature/Bug description:</b>  When the user creates a comment or reply, the name of his current Institution is not being send in this request.</p>

<p><b>Solution:</b> It was modified the post methods from PostCommentHandler and ReplyCommentHandler to get the user currentInstitution</p>

<p><b>TODO/FIXME:</b> n/a</p>


frontend/test/specs/postDetailsControllerSpec.js
```diff
@@ -51,6 +51,7 @@
         AuthService.getCurrentUser = function() {
             return user;
         };
+        commentService.user = user;
```
mayzabeel: precisa disso aqui?

backend/test/post_comment_handler_test.py
```diff
@@ -32,44 +32,79 @@ def setUp(cls):
...
+            }
+        }
 
-    @patch('utils.verify_token', return_value={'email': MAIANA_EMAIL})
-    def test_post(self, verify_token):
-        """Another user comment in Post of Mayza."""
+    @patch('handlers.post_comment_handler.enqueue_task')
+    @patch('utils.verify_token', return_value={'email': OTHER_USER_EMAIL})
+    def test_post(self, verify_token, enqueue_task):
+        """Another user comment in Post of user."""
```
andreldsa: Why another? If it's repeated maybe it's not necessary.

backend/test/reply_comment_handler_test.py
```diff
@@ -35,30 +36,81 @@ def setUp(cls):
...
+                self.third_user.key.urlsafe(),
+                "COMMENT",
+                self.user_post.key.urlsafe(),
+                body['currentInstitution']
+            )
+        ]
+        send_message_notification.assert_has_calls(calls)
 
         # Verify that the post is published
         self.assertEquals(self.user_post.state, "published")
```
andreldsa: The tests should end here, once, test a single scope of functionality, down below, it's testing delete post, that should be another test.

backend/test/post_comment_handler_test.py
ruanres: On test_post, i am testing the case a user, that is not the post author, comment on the post.
On test_post_ownerpost, i am testing the case in which the post's author comments on its own post.

PR #673, opened 2018-01-19
Improving search

raonismaneoto:
The search's filters in frontend weren't working as expected. 
</p>

<p><b>Solution:</b>
I've implemented a logic, manipulating the data, that makes the filters work at the same time.
</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/search/searchController.js
+++ b/frontend/search/searchController.js
@@ -53,7 +53,7 @@
         };
 
         searchCtrl.searchBy = function searchBy(search, attribute) {
-            if (searchInServer()) {
+            if (keywordHasChanges()) {
                 searchCtrl.makeSearch(search, 'institution', attribute);
                 refreshPreviousKeyword();
             } else {
@@ -128,7 +128,7 @@
             }
         }
 
-        function searchInServer() {
+        function keywordHasChanges() {
             var keywordHasChanged = searchCtrl.search_keyword != searchCtrl.previous_keyword;
             return _.isEmpty(searchCtrl.initialInstitutions) || keywordHasChanged || !searchCtrl.search_keyword;
         }
--- a/frontend/search/searchController.js
+++ b/frontend/search/searchController.js
@@ -52,11 +52,9 @@
 
         searchCtrl.searchBy = function searchBy(search, attribute) {
             if (searchInServer())  {
-                console.log("oi");
                 searchCtrl.makeSearch(search, 'institution', attribute);
                 searchCtrl.previous_keyword = searchCtrl.search_keyword;
             } else {
-                console.log("oi2");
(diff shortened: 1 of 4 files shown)


frontend/search/searchController.js
```diff
@@ -43,23 +50,55 @@
...
-                searchCtrl.actuationAreas = response.data;
+                searchCtrl.actuationAreas = objectToObjectArray(response.data);
+                actuationAreas = response.data;
             });
         }
 
         function getLegalNatures() {
             $http.get('app/institution/legal_nature.json').then(function success(response) {
-                searchCtrl.legalNature = response.data;
+                searchCtrl.legalNature = objectToObjectArray(response.data);
```
andreldsa: $http must be used only by a service.

frontend/search/searchController.js
```diff
@@ -73,6 +112,18 @@
             searchCtrl.brazilianFederalStates = brCidadesEstados.estados;
         }
 
+        function objectToObjectArray(object) {
```
andreldsa: https://lodash.com/docs/4.17.4#toArray

frontend/search/searchController.js
```diff
@@ -28,6 +34,7 @@
         searchCtrl.search = function search() {
             if (searchCtrl.search_keyword) {
                 searchCtrl.makeSearch(searchCtrl.search_keyword, 'institution');
+                searchCtrl.previous_keyword = searchCtrl.search_keyword;
```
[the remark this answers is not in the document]
raonismaneoto: The controller only calls the backend If the search's string change, otherwise the search is actually a filtering operation into the search result that already exists. So, those two variables controls If there is a change or not in the search's string.

frontend/search/searchController.js
```diff
@@ -72,16 +81,17 @@
...
             var stateIsNotSelected = !searchCtrl.searchState || stateDefaultValue;
 
-            // True If the fields are the same or If the field is not selected. 
+            // True If the institution's fields and controller's fields are the same or If the field is not selected. 
             var sameNature = legalNatures[institution.legal_nature] === searchCtrl.searchNature || natureIsNotSelected;
             var sameActuationArea = actuationAreas[institution.actuation_area] === searchCtrl.searchActuation || actuationIsNotSelected;
             var sameState = stateIsNotSelected || institution.federal_state === searchCtrl.searchState.nome;
 
-            // It keeps If the search's current attribute is the same in the institution and the search param and
+            // It stores If the search's current attribute is the same in the institution and in the search param and
```
andreldsa: If you separated this function with comment blocks, that's it's a sign that you have different contexts, and that should be another function well-documented.

PR #674, opened 2018-01-22
Fix show follow button in institution controller

Luiz-FS:
<p><b>Feature/Bug description:</b>  When entering the institution page several errors appear on the console indicating that the current instittution is null when the showFollowButton method is called.</p>

<p><b>Solution:</b> Add one more check in the showFollowButton method, which makes sure the current intitution exists before checking its properties.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/institution/institutionController.js
+++ b/frontend/institution/institutionController.js
@@ -136,7 +136,7 @@
         };
 
         institutionCtrl.showFollowButton = function showFollowButton() {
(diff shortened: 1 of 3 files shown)


frontend/institution/institutionController.js
```diff
@@ -9,7 +9,7 @@
         var morePosts = true;
         var actualPage = 0;
 
-        institutionCtrl.current_institution = null;
+        institutionCtrl.institution = null;
```
mayzabeel: Trocar no teste desse controller também

PR #684, opened 2018-01-23
Hide finalized events in home page

Luiz-FS:
<p><b>Feature/Bug description:</b> Hide preview events that are already finalized on the home page.</p>

<p><b>Solution:</b> I created a function in homeController that checks if the end date of the event has passed, if yes it does not let the event be viewed.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/home/homeController.js
+++ b/frontend/home/homeController.js
@@ -19,6 +19,7 @@
         homeCtrl.refreshTimeline = false;
         homeCtrl.instMenuExpanded = false;
         homeCtrl.isLoadingPosts = true;
+        homeCtrl.showMessageOfEmptyEvents = true;
         homeCtrl.stateView = "";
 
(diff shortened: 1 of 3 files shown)


frontend/home/homeController.js
```diff
@@ -35,6 +36,18 @@
             $state.go('app.institution.timeline', {institutionKey: institutionKey});
         };
 
+        homeCtrl.showEvent = function showEvent(event) {
```
mayzabeel: Teste

Luiz-FS: Done

andreldsa: showEvent seams to be an action that acts directly to render the event in the view. But, this is just a function that verifies if the event already happened.

PR #687, opened 2018-01-23
Update followers of health ministry

mayzabeel:
<p><b>Feature/Bug description:</b>When the new user follow the health ministry institution only the user was update.</p>

<p><b>Solution:</b> Update de Health Ministry too, when the new user follow</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/backend/utils.py
+++ b/backend/utils.py
@@ -122,10 +122,8 @@ def login_required(method):
     return login
 
 def follow_health_ministry(user,health_ministry):
-    user.follows.append(health_ministry.key)
-    user.put()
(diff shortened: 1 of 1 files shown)


backend/utils.py
```diff
@@ -121,6 +121,11 @@ def login(self, *args):
         method(self, user, *args)
     return login
 
+def follow_health_ministry(user,health_ministry):
+    user.follows.append(health_ministry.key)
```
andreldsa: You should use the functions of the model, user.follow(..) and institution.follow(...)

PR #688, opened 2018-01-23
Adding the links below the buttons in emails html

raonismaneoto:
The emails' buttons weren't clickable in hotmail.
</p>

<p><b>Solution:</b> 
I've added the links right below the buttons.

screenshot from 2018-01-23 16-14-51

</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/backend/templates/invite_institution_email.html
+++ b/backend/templates/invite_institution_email.html
@@ -21,17 +21,17 @@
                     </tr>
                 </table>
 
-                <table width="100%" border="0" cellpadding="0" cellspacing="0" align="center">
+                <table width="100%" height="100%" border="0" cellpadding="0" cellspacing="0" align="center">
                     <tr>
                         <td>
-                            <div style="height: 350px;">
+                            <div style="height: 480px;">
                                 <div style="background-color: #009688; height: 75%;" align="center">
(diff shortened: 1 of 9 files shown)


backend/templates/invite_institution_email.html
```diff
@@ -21,17 +21,17 @@
                     </tr>
                 </table>
 
-                <table width="100%" border="0" cellpadding="0" cellspacing="0" align="center">
+                <table width="100%" height="100%" border="0" cellpadding="0" cellspacing="0" align="center">
```
andreldsa: We have to be careful when expressing height in percentage, once, some browsers not allow that unit to style HTML. You should use another type of measurement or test the style with other browsers.

backend/templates/invite_institution_email.html
```diff
@@ -90,10 +96,10 @@
                     </tr>
                 </table>
 
-                <table width="100%" border="0" cellpadding="0" cellspacing="0" align="center" style="padding-bottom: 10px; margin-top: 100px;">
+                <table width="100%" height="100%" border="0" cellpadding="0" cellspacing="0" align="center" style="padding-bottom: 10px;">
```
andreldsa: same

PR #696, opened 2018-01-24
Fix ministério da saúde can't be removed

raonismaneoto:
The "Ministério da Saúde" can't be removed from the platform, but it was allowed before.
</p>

<p><b>Solution:</b>
I've added the name verification in frontend/backend.
</p>

<p><b>TODO/FIXME:</b> n/a</p>


frontend/institution/institutionController.js
```diff
@@ -267,16 +267,20 @@
...
-                templateUrl: 'app/institution/removeInstDialog.html',
-                targetEvent: ev,
-                clickOutsideToClose:true,
-                locals: {
-                    institution: institutionCtrl.institution
-                },
-                controller: RemoveInstController,
-                controllerAs: 'ctrl'
-            });
+            if (institutionCtrl.institution.name != "Ministério da Saúde") {
```
tiagolimpe: "!==" is more secure that "!="

andreldsa:
This not viable.

PR #701, opened 2018-01-24
Create Institution DECIIS

mayzabeel:
<p><b>Feature/Bug description:</b> The super institution now is DECIIS, so this institutuion must be create with MS, and the adm has permission to send invite for new institutions whe is logged with DECIIS, not more when is logged by MS. Other thing is that all user have to follow MS and DECIIS Institution</p>

- <p><b>Solution:</b>

- Create the new institution DECIIS
- Insert the official email of adm to deciis@saude.gov.br
- Force all user to follow DECIIS and MS institutions
</p>

<p><b>TODO/FIXME:</b> Think of a better way to do  make all users follow this institutions</p>

### DIFF
--- a/frontend/institution/institutionController.js
+++ b/frontend/institution/institutionController.js
@@ -139,7 +139,11 @@
         };
 
         institutionCtrl.showFollowButton = function showFollowButton() {
-           return institutionCtrl.institution && !institutionCtrl.isMember && institutionCtrl.institution.name !== "Ministério da Saúde";
+           return institutionCtrl.institution && 
+           !institutionCtrl.isMember && 
+           institutionCtrl.institution.name !== "Ministério da Saúde" &&
+           institutionCtrl.institution.name !== "Departamento do Complexo Industrial e Inovação em Saúde"
+           ;
         };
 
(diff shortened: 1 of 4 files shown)


frontend/institution/institutionController.js
```diff
@@ -139,7 +139,11 @@
         };
 
         institutionCtrl.showFollowButton = function showFollowButton() {
-           return institutionCtrl.institution && !institutionCtrl.isMember && institutionCtrl.institution.name !== "Ministério da Saúde";
+           return institutionCtrl.institution && 
+           !institutionCtrl.isMember && 
+           institutionCtrl.institution.name !== "Ministério da Saúde" &&
```
andreldsa: This is not a good approach, to minimize extract to an array of blocked institutions to unfollow, and use lodash, _.includes(array, insttitution_name), to verifies if it's part of the denied institutions.

andreldsa: Create a TODO to create a service that lead's with that logic of denied institutions, to prevent the spread of that verification turning the refactoring too much costly.

PR #704, opened 2018-01-25
Fix edit user

MaianaB:
<p><b>Feature/Bug description:</b> In edit user page when change the name or cpf withou click in button "Concluir edição" the change was saved.</p>

<p><b>Solution:</b> Created the copy of user that contain new information, when save edition the original user is changed..</p>

different

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/auth/configProfileController.js
+++ b/frontend/auth/configProfileController.js
@@ -127,8 +127,7 @@
                 controllerAs: "editProfileCtrl",
                 locals: {
                     institution: inst,
-                    user: configProfileCtrl.user,
-                    user: configProfileCtrl.newUser
+                    user: configProfileCtrl.user
                 },
                 targetEvent: ev,
                 clickOutsideToClose: true
--- a/frontend/auth/configProfileController.js
+++ b/frontend/auth/configProfileController.js
@@ -11,7 +11,8 @@
         // Variable used to observe the changes on the user model.
         var observer;
 
-        configProfileCtrl.newUser = AuthService.getCurrentUser();
+        configProfileCtrl.user = AuthService.getCurrentUser();
+        configProfileCtrl.newUser = _.clone(configProfileCtrl.user);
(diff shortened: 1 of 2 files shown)


frontend/auth/configProfileController.js
```diff
@@ -71,6 +72,7 @@
         function saveUser() {
             var deffered = $q.defer();
             if (configProfileCtrl.newUser.isValid()) {
+                changeUser();
```
raonismaneoto: updateUser would be better, wouldn't it?

frontend/auth/configProfileController.js
```diff
@@ -130,7 +139,7 @@
         }
 
         function hasMoreThanOneInstitution() {
-            return configProfileCtrl.newUser.institutions.length > 1;
+            return configProfileCtrl.user.institutions.length > 1;
```
raonismaneoto: Use _.size to keep the pattern

frontend/auth/configProfileController.js
```diff
@@ -144,11 +153,11 @@
         }
 
         function removeConection(institution_key) {
-            if (configProfileCtrl.newUser.institutions.length > 1) {
-                _.remove(configProfileCtrl.newUser.institutions, function(institution) {
+            if (configProfileCtrl.user.institutions.length > 1) {
```
raonismaneoto: Use _.size to keep the pattern

frontend/auth/configProfileController.js
```diff
@@ -158,7 +167,7 @@
         }
 
         function isAdminOfAnyInstitution() {
-            return configProfileCtrl.newUser.institutions_admin.length > 0;
+            return configProfileCtrl.user.institutions_admin.length > 0;
```
raonismaneoto: return !_.isEmpty(...)

PR #720, opened 2018-01-29
Add current institution name to invites and requests

ruanres:
<p><b>Feature/Bug description:</b> When the user rejects or accepts a received invite to create an institution or to become a member of one, the name of his current Institution is not being send in the request.</p>

<p><b>Solution:</b> It was modified some methods from models and handlers related to invites and requests to deal with the current institution data</p>

<p><b>TODO/FIXME:</b> 
Update another handlers related to requests to deal with the current institution data.
And add a verification method to check if the user belongs to the current institution passed in the request</p>


backend/test/institution_parent_request_handler_test.py
```diff
@@ -61,20 +65,28 @@ def test_put(self, verify_token, mock_method):
...
             exception_message,
             "Expected error message is Error! User is not allowed to accept link between institutions")
 
+    @patch('service_messages.send_message_notification')
     @patch('utils.verify_token', return_value={'email': 'otheruser@test.com'})
-    @mock.patch('service_messages.send_message_notification')
-    def test_delete(self, verify_token, mock_method):
+    def test_delete(self, verify_token, send_message_notification):
         """Test method post of InstitutionChildrenRequestHandler."""
+        print "under test"
```
mayzabeel: remover esses prints

PR #721, opened 2018-01-29
Fix bug of accept invite from email

Luiz-FS:
<p><b>Feature/Bug description:</b> Fix bug by accepting the invitation for the link received by email. When the user already logged in system and try to accept the invitation from the link received by email, it is redirected to a timeline and not to the invitation page as it should be.</p>

<p><b>Solution:</b> I created a method of name 'signin' in the acceptInviteController, which checks if the user is already logged in, if yes, redirects to the invitation page, otherwise goes to the login page.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/invites/acceptInviteController.js
+++ b/frontend/invites/acceptInviteController.js
@@ -23,10 +23,11 @@
         };
 
         controller.signin = function signin() {
-            if (AuthService.isLoggedIn())
+            if (AuthService.isLoggedIn()) {
                 $state.go("new_invite", {key: invite_id});
-            else
+            } else {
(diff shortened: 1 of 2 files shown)


frontend/invites/acceptInviteController.js
```diff
@@ -21,6 +21,13 @@
         controller.goToHome = function goToHome() {
             $state.go("app.user.home");
         };
+
+        controller.signin = function signin() {
+            if (AuthService.isLoggedIn())
```
andreldsa: use {} to organize if else statements.

PR #722, opened 2018-01-30
Fix edit profile's image in mozilla

raonismaneoto:
It wasn't possible to edit the profile's image in mozilla because the browser doesn't work with percentage in the height property and It was missing a $event in the html.
</p>

<p><b>Solution:</b>
I've changed the height property from percentage to vh and added $event where it was missing.

screenshot from 2018-01-30 08-56-25

</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/auth/configProfileController.js
+++ b/frontend/auth/configProfileController.js
@@ -47,8 +47,8 @@
             });
         }
 
-        configProfileCtrl.cropImage = function cropImage(imageFile) {
-            CropImageService.crop(imageFile).then(function success(croppedImage) {
+        configProfileCtrl.cropImage = function cropImage(imageFile, event) {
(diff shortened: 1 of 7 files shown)


frontend/auth/configProfileController.js
```diff
@@ -47,8 +47,8 @@
             });
         }
 
-        configProfileCtrl.cropImage = function cropImage(imageFile) {
-            CropImageService.crop(imageFile).then(function success(croppedImage) {
+        configProfileCtrl.cropImage = function cropImage(imageFile, event) {
```
mayzabeel: Mesmo tratamento deve ser feito ao trocar a imagem da instuição em ConfigInstController

andreldsa: Fix then it's ok

PR #731, opened 2018-01-31
Adding personalized scrollbar in screens of system.

tiagolimpe:
<p><b>Feature/Bug description:</b> In timeline and other screens that has scroll, the display option of scroll is None.</p>

<p><b>Solution:</b> Adding a personalized scrollbar in screens that has scroll.</p>

screenshot from 2018-01-30 16-07-12

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/styles/custom.css
+++ b/frontend/styles/custom.css
@@ -523,16 +523,19 @@ a:active {
 }
 
 .hide-scrollbar::-webkit-scrollbar {
+    display: none;
+}
+.custom-scrollbar::-webkit-scrollbar {
     width: .3em;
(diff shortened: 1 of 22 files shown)


frontend/styles/custom.css
```diff
@@ -523,7 +523,17 @@ a:active {
 }
 
 .hide-scrollbar::-webkit-scrollbar {
-    display: none;
+    width: .3em;
```
mayzabeel: trocar o nome hide-scrollbar por custom-scrollbar por exemplo

PR #749, opened 2018-02-02
Block request institution button

raonismaneoto:
The request institution's button was in institution's page, with the same message and clickable, even if the user had already requested to be a member.
</p>

<p><b>Solution:</b>
I've created a handler that allows to get the user's requests and got these requests in institutionController. It allowed the controller to know if the user has requested to be a member.

screenshot from 2018-02-02 14-44-51

</p>

<p><b>TODO/FIXME:</b> n/a</p>


mayzabeel:
Closed because is too expensive to only  verify a button

PR #768, opened 2018-02-07
Improve post of vote handler

mayzabeel:
<p><b>Feature/Bug description:</b>Refact to minimize concurrency problems</p>

<p><b>Solution:</b>isolate only the transactional action and update to try 10 times</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/backend/models/survey_post.py
+++ b/backend/models/survey_post.py
@@ -68,6 +68,7 @@ class SurveyPost(Post):
 
         return True
 
+    @ndb.transactional(retries=10)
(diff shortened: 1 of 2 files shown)


backend/models/survey_post.py
```diff
@@ -77,6 +78,8 @@ def vote(self, author, all_options_selected):
             for option in all_options_selected:
                 self.add_vote(author, option["id"])
```
andreldsa: RLB

PR #799, opened 2018-02-19
App selection before deploy through ecis script

andreldsa:
<p><b>Feature/Bug description:</b> Add an option to select which application it will be deployed.</p>

<p><b>Solution:</b> Add an input reader to the script, attaching that to a variable modifying the domain name and application project of GCloud configuration.

screenshot from 2018-02-19 14-15-15
</p>

<p><b>TODO/FIXME:</b> n/a</p>


ecis
```diff
@@ -112,7 +108,26 @@ case "$1" in
     ;;
 
     deploy)
-        echo "=========== Staring Google App Engine Deployment ==========="
+        read -p "${bold}>> Type the app name:${_bold} " APP_NAME
```
mayzabeel: Massa, era bom ter comentado todos os APP_NAMEs que temos hoje em algum lugar!

PR #801, opened 2018-02-19
Fix reply deleted comment

raonismaneoto:
The reply_comment method wasn't checking if the comment really exists before add the reply.
</p>

<p><b>Solution:</b>
I've added a verification in the method and sent a message to the user warning him that the comment has been deleted.

screenshot from 2018-02-19 14-56-18

</p>

<p><b>TODO/FIXME:</b> n/a</p>


backend/models/post.py
```diff
@@ -252,7 +253,9 @@ def remove_comment(self, comment):
         self.put()
 
     def reply_comment(self, reply, comment_id):
-        comment = self.comments.get(comment_id)
+        comment = self.get_comment(comment_id)
```
mayzabeel: teste

PR #804, opened 2018-02-19
Display the result of a survey after deadline

tiagolimpe:
<p><b>Feature/Bug description:</b> The result of survey is only displayed in case of user is author or has voted in survey.</p>

<p><b>Solution:</b> Now the user can see the result of survey after deadline.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/survey/surveyDetailsDirective.js
+++ b/frontend/survey/surveyDetailsDirective.js
@@ -40,7 +40,13 @@
 
         surveyCtrl.isSurveyAuthor = function isSurveyAuthor() {
             return surveyCtrl.post.author_key == surveyCtrl.user.key;
-        }
+        };
+
+        surveyCtrl.timeHasBeenExpired = function timeHasBeenExpired() {
+            var nationalTimeZone = new Date (surveyCtrl.post.deadline);
+            nationalTimeZone.setHours(nationalTimeZone.getHours() - 3);
+            return new Date() > nationalTimeZone;
+        };
 
         surveyCtrl.canVote = function(){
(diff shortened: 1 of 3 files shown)


frontend/survey/surveyDetailsDirective.js
```diff
@@ -40,7 +40,13 @@
 
         surveyCtrl.isSurveyAuthor = function isSurveyAuthor() {
             return surveyCtrl.post.author_key == surveyCtrl.user.key;
-        }
+        };
+
+        surveyCtrl.timeHasBeenExpired = function timeHasBeenExpired() {
+            var nationalTimeZone = new Date (surveyCtrl.post.deadline);
+            nationalTimeZone.setHours(nationalTimeZone.getHours() - 3);
```
andreldsa: That subtraction to get a timezone can cause errors, once that 3 is not valid to all users, think how that is gone to work if the user its placed outside of Brazil, or in the -4 Brazilian timezone (acre). You should work with momentJs, that we already have imported.

frontend/survey/surveyDetailsDirective.js
```diff
@@ -40,7 +40,13 @@
 
         surveyCtrl.isSurveyAuthor = function isSurveyAuthor() {
             return surveyCtrl.post.author_key == surveyCtrl.user.key;
-        }
+        };
+
+        surveyCtrl.timeHasBeenExpired = function timeHasBeenExpired() {
```
andreldsa: Test

PR #814, opened 2018-02-20
Fixed responsideness of event page

MaianaB:
<p><b>Feature/Bug description:</b> Parts of page was out of card.</p>

<p><b>Solution:</b> Put layout-row in small displays and "flex" has only in display bigger than 600.</p>

<p><b>TODO/FIXME:</b> n/a</p>

screenshot from 2018-02-20 11-08-48
screenshot from 2018-02-20 11-08-42


andreldsa:
@MaianaB That view should be flex 100 on mobile screens.

PR #816, opened 2018-02-20
Refact method POST of  like handler

mayzabeel:
<p><b>Feature/Bug description:</b> Optimize method post to like post, comment or reply to minimize concurrency problems</p>

<p><b>Solution:</b> Isolate the transsaction only to part methods that update the datastore  and insert the flag retries=10 to try more than once on error cases </p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/backend/handlers/like_handler.py
+++ b/backend/handlers/like_handler.py
@@ -46,19 +46,15 @@ class LikeHandler(BaseHandler):
     @login_required
     def post(self, user, post_key, comment_id=None, reply_id=None):
         """Handle POST Requests."""
-        """This method is only meant to give like in post, comment or reply."""
+        """This method is only meant to give like in post, comment or reply and send notification."""
         if comment_id:
             @ndb.transactional(retries=10)
             def like_comment(post_key, user, comment_id=None, reply_id=None):
-                """Increment one 'like' in  comment or reply and send notification.""" 
+                """Increment one 'like' in  comment or reply""" 
                 post = ndb.Key(urlsafe=post_key).get()
-                comment = post.get_comment(comment_id)
-                if reply_id:
-                    comment = comment.get('replies').get(reply_id)
-                post.like_comment(comment, user.key.urlsafe() )
+                update_comment = post.like_comment(user, comment_id, reply_id)
+                return update_comment
                 
-                return comment
-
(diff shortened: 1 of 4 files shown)


backend/handlers/like_handler.py
```diff
@@ -44,54 +44,62 @@ def get(self, user, post_key, comment_id=None, reply_id=None):
...
+            def like_comment(post_key, user, comment_id=None, reply_id=None):
+                """Increment one 'like' in  comment or reply and send notification.""" 
+                post = ndb.Key(urlsafe=post_key).get()
+                comment = post.get_comment(comment_id)
+                if reply_id:
+                    comment = comment.get('replies').get(reply_id)
+                likes = comment.get('likes')
+                Utils._assert(user.key.urlsafe() in likes,
+                            "User already liked this comment", NotAuthorizedException)
+                likes.append(user.key.urlsafe())
```
andreldsa: You should extract that list modification to the model, avoiding the handler to modify the entities.

backend/handlers/like_handler.py
```diff
@@ -44,54 +44,62 @@ def get(self, user, post_key, comment_id=None, reply_id=None):
...
+            def like(post_key, user):
+                post = ndb.Key(urlsafe=post_key).get()
+                Utils._assert(post.key in user.liked_posts, 
+                    "User already liked this publication", NotAuthorizedException)
+                user.like_post(post.key)
+                if post.get_like(user.key) is None:
+                    like = Like()
+                    like.author = user.key
+                    like.id = Utils.getHash(like)
+                    post.likes.append(like)
```
andreldsa: same

backend/handlers/like_handler.py
```diff
@@ -44,54 +44,54 @@ def get(self, user, post_key, comment_id=None, reply_id=None):
...
-            if reply_id:
-                comment = comment.get('replies').get(reply_id)
-
-            likes = comment.get('likes')
-
-            Utils._assert(user.key.urlsafe() in likes,
-                      "User already liked this comment", NotAuthorizedException)
-            likes.append(user.key.urlsafe())
-            post.put()
+            @ndb.transactional(retries=10)
```
andreldsa: Extract the content of those if else to external functions.

PR #821, opened 2018-02-22
Fix event changes in the edit dialog that directly modified the event page.

MaianaB:
<p><b>Feature/Bug description:</b> Any modifications made to the event through the edit event dialog is modified to the page, even if the user cancels the edition. This error is only in frontend, the changes weren't saved in backend.</p>

<p><b>Solution:</b> Passed the copy of event to edit controller, and when user save the modifies is returned the new object.</p>

<p><b>TODO/FIXME:</b> n/a</p>

### DIFF
--- a/frontend/event/eventController.js
+++ b/frontend/event/eventController.js
@@ -404,16 +404,9 @@
         }
 
         function loadEventDates() {
-            dialogCtrl.start_time = toLocaleDate(dialogCtrl.dateChangeEvent.start_time);
-            //dialogCtrl.event.start_time = toLocaleDate(dialogCtrl.dateChangeEvent.start_time);
-            //dialogCtrl.event.end_time = toLocaleDate(dialogCtrl.dateChangeEvent.end_time);
-        }
-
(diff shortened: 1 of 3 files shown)


frontend/event/eventController.js
```diff
@@ -404,9 +404,9 @@
         }
 
         function loadEventDates() {
-            dialogCtrl.start_time = new Date(dialogCtrl.dateChangeEvent.start_time);
-            dialogCtrl.event.start_time = new Date(dialogCtrl.dateChangeEvent.start_time);
-            dialogCtrl.event.end_time = new Date(dialogCtrl.dateChangeEvent.end_time);
+            dialogCtrl.start_time = moment.utc(dialogCtrl.dateChangeEvent.start_time)._d;
```
andreldsa: why the property _d?

MaianaB: Foi um exemplo que vi na internet. Mas já tô pensando em outra forma, pois vi na documentação que não deve ser usada.

PR #830, opened 2018-02-23
Fix reload of survey

tiagolimpe:
<p><b>Feature/Bug description:</b> When the survey is reloaded the voted option and the percentage disappear.</p>

<p><b>Solution:</b> Changing the type_survey to false to reload attributes when the response is finished.</p>

<p><b>TODO/FIXME:</b> n/a</p>


frontend/post/postDetailsDirective.js
```diff
@@ -312,11 +312,15 @@
         };
 
         postDetailsCtrl.reloadPost = function reloadPost() {
+            const type_survey = postDetailsCtrl.post.type_survey;
```
andreldsa: Let's use var, to minimize problems with compatibility in the near future.

PR #832, opened 2018-02-23
Fix create event even when the user don't set de event date

MaianaB:
<p><b>Feature/Bug description:</b> User can pass to next step even if he didn't put the date.</p>

<p><b>Solution:</b> Added verification when go to next step.</p>

<p><b>TODO/FIXME:</b> n/a</p>

screenshot from 2018-02-23 13-05-45
screenshot from 2018-02-23 13-05-30


frontend/event/eventController.js
```diff
@@ -281,6 +281,16 @@
             return valid;
         };
 
+        dialogCtrl.isValidDate = function isValidDate(){  
+            var isUndefined = _.isUndefined(dialogCtrl.event.start_time) ||
```
andreldsa: It doesn't need the variable.

frontend/event/eventController.js
```diff
@@ -281,6 +281,16 @@
             return valid;
         };
 
+        dialogCtrl.isValidDate = function isValidDate(){
```
andreldsa: add some tests.

frontend/event/eventController.js
```diff
@@ -282,12 +282,11 @@
         };
 
         dialogCtrl.isValidDate = function isValidDate(){  
-            var isUndefined = _.isUndefined(dialogCtrl.event.start_time) || 
-                _.isUndefined(dialogCtrl.event.end_time);
-            return !isUndefined;
+            return  !_.isUndefined(dialogCtrl.event.start_time) ||
```
andreldsa: The negative form of an "or" statement is "not a AND not b".

frontend/test/specs/eventDialogCtrlSpec.js
```diff
@@ -267,19 +267,41 @@
...
+          controller.event.title = "Title";
+          controller.event.address = address;
+          controller.event.local = "Local";
+          controller.event.start_time = undefined;
+          controller.event.end_time = undefined;
+          controller.nextStep();
+          expect(controller.steps).toEqual([true, false, false]);
+        });
+
+        it('should pass from first and second step', function() {
```
andreldsa: You should create a test that covers the problem spoken above. In your new logic 
```
controller.event.start_time = undefined;
controller.event.end_time = date_now;
```

Will work, but shouldn't.

PR #845, opened 2018-02-27
Request user confirmation to leave the profile editing page without saving changes.

MaianaB:
<p><b>Feature/Bug description:</b> When the user leaves the edit page while editing the profile, he loses the changes and he is not warned about it. </p>

<p><b>Solution:</b> Added directive that confirm this action.</p>

<p><b>TODO/FIXME:</b> n/a</p>

screenshot from 2018-02-27 13-29-04

### DIFF
--- a/frontend/auth/configProfileController.js
+++ b/frontend/auth/configProfileController.js
@@ -3,7 +3,7 @@
 (function () {
     var app = angular.module("app");
 
-    app.controller("ConfigProfileController", function ConfigProfileController($state, InstitutionService,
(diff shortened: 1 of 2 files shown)


frontend/auth/configProfileController.js
```diff
@@ -3,7 +3,7 @@
 (function () {
     var app = angular.module("app");
 
-    app.controller("ConfigProfileController", function ConfigProfileController($state, InstitutionService,
+    app.controller("ConfigProfileController", function ConfigProfileController($state, InstitutionService, SubmitFormListenerService,
```
andreldsa: Why inject that service? You don't need it to be able to use the directive.

PR #854, opened 2018-03-01
Fix send notification in post sharing

raonismaneoto:
The sharing post's notification was being sent to the wrong receiver and with a wrong body.
Besides, the iteration for send followers notification wasn't in a queue.
</p>

<p><b>Solution:</b>
I've added the iteration in a queue and fixed the sharing post notification.

screenshot from 2018-03-01 11-08-03

</p>

<p><b>TODO/FIXME:</b> n/a</p>


backend/handlers/post_collection_handler.py
```diff
@@ -72,26 +72,38 @@ def create_post(post_data, user, institution):
...
             entity_type = 'SHARED_POST'
             params = {
-                'receiver_key': post.author.urlsafe(),
+                'receiver_key': shared_post.author.urlsafe(),
                 'sender_key': user.key.urlsafe(),
                 'entity_key': post.key.urlsafe(),
                 'entity_type': entity_type,
-                'current_institution': user.current_institution.urlsafe()
+                'current_institution': user.current_institution.urlsafe(),
+                'subscribers': json.dumps([subscriber.urlsafe() for subscriber in shared_post.subscribers])
```
andreldsa: This "for" should not be done in the handler, subscribers list could lead to the whole number of users of the system and can take too much time to compute.

PR #868, opened 2018-03-05
Search registered institutions

tiagolimpe:
**Feature/Bug description:** Now it's possible to filter institutions by a keyword in registered institutions page (all institutions).

**Solution:** Creating a field to search by keyword, adding a filter in ng-repeat and function in Utils to normalize the strings (removing accents and special chars).

screenshot from 2018-03-05 16-57-08

screenshot from 2018-03-05 16-57-31

screenshot from 2018-03-05 16-58-04

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/institution/institutions.html
+++ b/frontend/institution/institutions.html
@@ -12,7 +12,7 @@
                     </div>
                 </md-toolbar>
                 <md-card-content>
-                    <div style="margin-top: 5%;">
+                    <div style="margin-top: 15px;">
                         <form>
                             <md-input-container class="md-block">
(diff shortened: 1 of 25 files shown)


frontend/institution/institutions.html
```diff
@@ -12,11 +12,19 @@ <h1>
                     </div>
                 </md-toolbar>
                 <md-card-content>
+                    <div style="margin-top: 5%;">
```
andreldsa: This percentage in a vertical perspective will cause layout malfunction. Use a fixed length.

PR #869, opened 2018-03-05
Fix invite status verification

raonismaneoto:
**Feature/Bug description:**
In some handlers the invite's status verification was wrong. It was checking if the invite wasn't accepted but It should check if it was sent.
**Solution:**
Change the verification and add it when it was missing.

**TODO/FIXME:** n/a


backend/handlers/invite_handler.py
```diff
@@ -76,8 +81,8 @@ def patch(self, user, invite_key):
         data = self.request.body
         invite = ndb.Key(urlsafe=invite_key).get()
 
-        Utils._assert(invite.status == 'accepted', 
-            "This invitation has already been accepted", 
+        Utils._assert(invite.status != 'sent',
```
mayzabeel: teste

PR #870, opened 2018-03-05
Fix the layout of a closed shared survey

MaianaB:
**Feature/Bug description:**  The layout of shared survey post is the same when the survey was closed or not.
**Solution:** Changed layout according to closed survey's layout.

**TODO/FIXME:** n/a

screenshot from 2018-03-05 15-51-08

### DIFF
--- a/frontend/post/postDetailsDirective.js
+++ b/frontend/post/postDetailsDirective.js
@@ -79,10 +79,10 @@
                 postDetailsCtrl.post.shared_post.type_survey;
         };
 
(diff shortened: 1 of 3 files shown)


frontend/post/postDetailsDirective.js
```diff
@@ -104,6 +104,11 @@
             return !hasNoComments || !hasNoLikes;
         };
 
+        postDetailsCtrl.isSharedSurveyExpired = function isSharedSurveyExpired() {
```
andreldsa: Test

PR #878, opened 2018-03-08
Request invitation button on institution card

ruanres:
**Feature/Bug description:** On institution card, at registered institution page, it was not possible to the user to request an invitation to any of those institutions.

**Solution:** It was created a button to the user request an invitation when he is not a member.

**TODO/FIXME:** n/a

request

### DIFF
--- a/backend/models/institution.py
+++ b/backend/models/institution.py
@@ -263,14 +263,13 @@ class Institution(ndb.Model):
 
     def make(self, attributes):
         """Create an institution dictionary with specific filds."""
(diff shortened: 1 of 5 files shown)


backend/models/institution.py
```diff
@@ -263,14 +263,13 @@ def change_state(self, state):
 
     def make(self, attributes):
         """Create an institution dictionary with specific filds."""
+        # import pdb
```
andreldsa: ??

PR #879, opened 2018-03-08
Show all notifications in notification page

MaianaB:
**Feature/Bug description:** In notification page show only notifications that weren't read by the user.

**Solution:** Get firebase component that has all notifications.

**TODO/FIXME:** n/a

screenshot from 2018-03-08 11-56-08


frontend/notification/notificationDirective.js
```diff
@@ -200,6 +201,11 @@
             });
         }
 
+        notificationCtrl.showNotifications = function showNotifications($mdMenu, $event) {
```
andreldsa: test

PR #882, opened 2018-03-09
Fix inconsistency of comment and reply comment 

MaianaB:
**Feature/Bug description:** When a user that was commented or reply comment in the post and then changed the photo or/and name the comment didn't update.

**Solution:**  When load comment, load attributes the author.

**TODO/FIXME:** n/a

screenshot from 2018-03-09 11-40-52
screenshot from 2018-03-09 11-40-05

### DIFF
--- a/backend/models/post.py
+++ b/backend/models/post.py
@@ -11,7 +11,6 @@ from utils import Utils
 import datetime
 
 
-
 def getCommentsUri(post, host):
     """Create uri to access post comments."""
     return "http://%s/api/posts/%s/comments" % (host, post.key.urlsafe())
--- a/backend/models/post.py
+++ b/backend/models/post.py
@@ -9,7 +9,6 @@ from models.event import Event
 from utils import Utils
 
 import datetime
-import json
 
 
 
--- a/backend/models/post.py
+++ b/backend/models/post.py
@@ -9,6 +9,8 @@ from models.event import Event
 from utils import Utils
 
 import datetime
+import json
+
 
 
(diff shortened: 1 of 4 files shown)


frontend/post/postDetailsDirective.js
```diff
@@ -706,7 +707,26 @@
...
 
-        loadShowReplies();
+        function updateComment(){
+            loadAuthor(commentCtrl.comment);
+            _.forEach(commentCtrl.comment.replies, function(value, key) {
+                loadAuthor(value);
+            });
+        }
+
+        function loadAuthor(comment) {
```
mayzabeel: Essa alteração deve ser feita no get de comentários no servidor, fazer uma requisição a cada comentário do lado cliente é mais custoso, melhor processar isso no servidor logo

frontend/post/postDetailsDirective.js
```diff
@@ -505,7 +505,8 @@
         };
     });
 
-    app.controller('CommentController', function CommentController(CommentService, MessageService, ProfileService, $state, AuthService) {
+    app.controller('CommentController', function CommentController(CommentService, MessageService, ProfileService, $state, 
+            AuthService, UserService) {
```
mayzabeel: remover userService

backend/models/post.py
```diff
@@ -11,6 +11,7 @@
 import datetime
 
 
+
```
mayzabeel: rlb

PR #886, opened 2018-03-13
Fix remove institution from the user

raonismaneoto:
**Feature/Bug description:** 
When the user remove an institution his current_institution wasn't changing what could create a bug.
**Solution:**
I've called removeInstitution and removeProfile in the removeConection method. Thus, the current_institution is changed once the removeInstitution method call changeInstitution().

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/auth/configProfileController.js
+++ b/frontend/auth/configProfileController.js
@@ -159,9 +159,7 @@
         function removeConection(institution_key) {
             if (_.size(configProfileCtrl.user.institutions) > 1) {
                 configProfileCtrl.user.removeInstitution(institution_key);
-                _.remove(configProfileCtrl.user.institution_profiles, function(profile) {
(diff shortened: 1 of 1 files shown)


frontend/auth/configProfileController.js
```diff
@@ -158,12 +158,8 @@
 
         function removeConection(institution_key) {
             if (_.size(configProfileCtrl.user.institutions) > 1) {
-                _.remove(configProfileCtrl.user.institutions, function(institution) {
-                    return institution.key === institution_key;
-                });
-                _.remove(configProfileCtrl.user.institution_profiles, function(profile) {
-                    return profile.institution_key === institution_key;
-                });
+                configProfileCtrl.user.removeInstitution(institution_key);
```
mayzabeel: essas duas funções tem teste? Se não tiver, fazer.

PR #897, opened 2018-03-15
Fix remove account after delete an institution

raonismaneoto:
**Feature/Bug description:**
The frontend wasn't updating the user.institutions_admin when deleting an institution. So, when the admin removed an institution he couldn't remove his account yet, because the inst was still there. Besides, the system wasn't send 'DELETED_INSTITUTION' notification because the worker had a little bug. Lastly, the patch object in the configProfileController, I don't know why, was getting the institution's remove operation.
**Solution:**
I've removed the inst from institutions_admin in the frontend, fixed the worker and make a patch without any observer.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/models/user.py
+++ b/backend/models/user.py
@@ -279,7 +279,7 @@ class User(ndb.Model):
     
     def remove_institution_admin(self, institution_key):
         """Remove a institution admin to user."""
-        if institution_key not in self.institutions_admin:
+        if institution_key in self.institutions_admin:
             self.institutions_admin.remove(institution_key)
             self.put()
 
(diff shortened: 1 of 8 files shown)


frontend/user/user.js
```diff
@@ -101,6 +101,16 @@ User.prototype.removeInstitution = function removeInstitution(institutionKey, re
     _.remove(this.institutions, toRemove);
     _.remove(this.follows, toRemove);
 
+    var isAdmin = _.find(this.institutions_admin, function (each) {
```
mayzabeel: Usar função já existente que checa isso user.isAdmin

backend/models/user.py
```diff
@@ -163,6 +163,7 @@ def remove_institution(self, institution):
             self.institutions.remove(institution)
             self.remove_permission('publish_post', institution.urlsafe())
             self.remove_profile(institution.urlsafe())
+            self.follows.remove(institution)
```
mayzabeel: acho que não precisa remover quem ele segue

PR #898, opened 2018-03-16
Fix Home Controller Test

mayzabeel:
**Feature/Bug description:** The test loadMorePosts in Hone Controller test is broken on master repository, the error is : permission_denied at /notifications/undefined: Client doesn't have permission to access the desired data.

**Solution:** The main controller function, has a request for NotificationService, and this call is not mocked on test.

**TODO/FIXME:**

### DIFF
--- a/frontend/test/specs/homeControllerSpec.js
+++ b/frontend/test/specs/homeControllerSpec.js
@@ -2,7 +2,7 @@
 
 (describe('Test HomeController', function() {
 
-    var homeCtrl, httpBackend, scope, createCtrl, mdDialog, state, postService, http;
+    var homeCtrl, httpBackend, scope, createCtrl, mdDialog, state, postService, http, notificationService;
 
     var institutions = [{
         acronym: 'Certbio',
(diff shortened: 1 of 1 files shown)


frontend/test/specs/homeControllerSpec.js
```diff
@@ -136,8 +137,21 @@
...
+                spyOn(notificationService, 'watchPostNotification').and.callFake(function() {
+                    return {
+                        then: function(callback) {
+                            callback();
+                        }
+                    };
+                });
             });
 
+            /*
```
Luiz-FS: Remove Fix comment

frontend/test/specs/homeControllerSpec.js
```diff
@@ -34,13 +34,14 @@
...
     beforeEach(inject(function($controller, $httpBackend, $rootScope, $q, InstitutionService,
-            PostService, $mdDialog, $state, AuthService, $http) {
+            PostService, $mdDialog, $state, AuthService, $http, NotificationService) {
         httpBackend = $httpBackend;
         http = $http;
         scope = $rootScope.$new();
         mdDialog = $mdDialog;
         state = $state;
         postService = PostService;
+        notificationService = NotificationService;
```
Luiz-FS: If the notificationsService mock is only being used within the beforeEach, you do not need to create a variable to access in the external scope because it already works for all tests.

PR #901, opened 2018-03-19
Fix console error in read of property 'cover_photo'

tiagolimpe:
**Feature/Bug description:** When loading the institution page the following console error is showed:
"Cannot read property 'cover_photo' of null".

**Solution:** Verifying if the institution is defined before reading the property 'cover_photo'.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/institution/institutionController.js
+++ b/frontend/institution/institutionController.js
@@ -133,7 +133,8 @@
         };
 
         institutionCtrl.showImageCover = function showImageCover(){
-            return !institutionCtrl.isLoadingCover && institutionCtrl.institution.cover_photo;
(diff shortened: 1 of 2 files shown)


frontend/institution/institutionController.js
```diff
@@ -133,7 +133,8 @@
         };
 
         institutionCtrl.showImageCover = function showImageCover(){
-            return !institutionCtrl.isLoadingCover && institutionCtrl.institution.cover_photo;
+            if(institutionCtrl.institution)
```
ruanres: colocar essa condição no return, antes de acessar cover_photo

PR #911, opened 2018-03-21
Fix accepted invite when institution is inactive

MaianaB:
**Feature/Bug description:** The user can accept the invitation that was sent by the institution that is currently inactive.

**Solution:** Added verification in handler so the user can't accept the invite and show only valid invites.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/handlers/invite_handler.py
+++ b/backend/handlers/invite_handler.py
@@ -26,6 +26,10 @@ def check_if_user_is_member(user, institution):
     """Check if the user is already a member."""
     return institution.has_member(user.key) and user.is_member(institution.key)
 
+def check_if_inst_is_active(institution):
+    """Check if the institution is active."""
+    return institution.state == "active"
+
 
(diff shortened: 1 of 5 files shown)


backend/search_module/search_institution.py
```diff
@@ -13,6 +13,7 @@ def institution_has_changes(fields, entity):
         if not isinstance(entity, Address):
             for field in fields:
                 if not hasattr(entity, field.name):
+                    print entity
```
mayzabeel: remove

backend/handlers/invite_handler.py
```diff
@@ -26,6 +26,10 @@ def check_if_user_is_member(user, institution):
     """Check if the user is already a member."""
     return institution.has_member(user.key) and user.is_member(institution.key)
 
+def check_if_inst_is_active(institution):
```
ruanres: Esse método deveria estar em institution.

PR #913, opened 2018-03-22
Show message when click message 

MaianaB:
**Feature/Bug description:** When the user clicks on invitation notification that was used show quickly the dialog and toast that invite has been resolved by the user.

**Solution:** Show the message in the dialog.

**TODO/FIXME:** n/a

screenshot from 2018-03-22 11-30-32
screenshot from 2018-03-22 11-30-07

### DIFF
--- a/frontend/requests/requestInstitutionProcessingController.js
+++ b/frontend/requests/requestInstitutionProcessingController.js
@@ -50,6 +50,10 @@
             $mdDialog.cancel();
         };
 
+        requestController.sizeGtSmDialog = function sizeGtSmDialog() {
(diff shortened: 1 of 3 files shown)


frontend/requests/requestInstitutionProcessingController.js
```diff
@@ -50,6 +50,10 @@
             $mdDialog.cancel();
         };
 
+        requestController.sizeGtSmDialog = function sizeGtSmDialog() {
+            return (requestController.request.status === 'sent') ? '45' : '25';
```
ruanres: não precisa dos parênteses

PR #914, opened 2018-03-23
Fix resend invite

MaianaB:
**Feature/Bug description:** When the user sends an invitation and then clicks to resend it, it is not possible because the invitation does not yet have the key.

**Solution:** When the invitations are made the post has returned an array that contains key and email objects

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/institution/managementMembersController.js
+++ b/frontend/institution/managementMembersController.js
@@ -100,7 +100,7 @@
                 manageMemberCtrl.isLoadingInvite = true;
                 var promise = InviteService.sendInvite(requestBody);
                 promise.then(function success(response) {
-                    refreshSentInvitations(requestBody.emails);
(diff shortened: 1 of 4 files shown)


frontend/institution/managementMembersController.js
```diff
@@ -114,10 +114,11 @@
             }
         };
 
-        function refreshSentInvitations(emails) {
+        function refreshSentInvitations(emails, invites) {
             var inviteToAdd = new Invite(manageMemberCtrl.invite);
             _.each(emails, function(email) {
                 inviteToAdd.invitee = email;
+                inviteToAdd.key = invites.reduce(invite => invite.email === email).key;
```
Luiz-FS: Set inicialization of reduce.

PR #915, opened 2018-03-26
Refact user handler

raonismaneoto:
**Feature/Bug description:** The user handler http methods wasn't cohesive.

**Solution:**
I've refactored the patch and delete methods and created another handler that wraps the logic of delete an institution from an user.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/handlers/user_handler.py
+++ b/backend/handlers/user_handler.py
@@ -85,7 +85,7 @@ class UserHandler(BaseHandler):
 
     @json_response
     @login_required
-    def delete(self, user, institution_key):
+    def delete(self, user):
         """Handler DELETE Requests."""
         user.state = 'inactive'
 
@@ -93,6 +93,8 @@ class UserHandler(BaseHandler):
         user.disable_account()
 
         user.put()
+
+        self.response.write(json.dumps(make_user(user, self.request)))
 
     @json_response
     @login_required
--- a/backend/handlers/user_handler.py
+++ b/backend/handlers/user_handler.py
@@ -87,26 +87,21 @@ class UserHandler(BaseHandler):
     @login_required
     def delete(self, user, institution_key):
         """Handler DELETE Requests."""
-        institution_key = ndb.Key(urlsafe=institution_key)
-        institution = institution_key.get()
+        user.state = 'inactive'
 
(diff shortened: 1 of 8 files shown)


backend/handlers/user_handler.py
```diff
@@ -85,15 +85,14 @@ def get(self, user):
 
     @json_response
     @login_required
-    def delete(self, user, institution_key):
+    def delete(self, user):
```
mayzabeel: Document the changed handler methods.

backend/handlers/user_institutions_handler.py
```diff
@@ -0,0 +1,44 @@
...
+from handlers.base_handler import BaseHandler
+
+
+class UserInstitutionsHandler(BaseHandler):
+    """Handle user's operations relationed to an especific institution."""
+
+    @json_response
+    @login_required
+    def delete(self, user, institution_key):
+        """Handle DELETE Requests."""
```
mayzabeel: Here too

backend/handlers/user_institutions_handler.py
```diff
@@ -0,0 +1,44 @@
...
+from utils import login_required
+from utils import Utils
+from utils import json_response
+from send_email_hierarchy.leave_institution_email_sender import LeaveInstitutionEmailSender
+
+from handlers.base_handler import BaseHandler
+
+
+class UserInstitutionsHandler(BaseHandler):
+    """Handle user's operations relationed to an especific institution."""
```
mayzabeel: if is specific institution why userInstitutioS?

backend/handlers/user_handler.py
```diff
@@ -85,15 +85,14 @@ def get(self, user):
...
         """Handler DELETE Requests."""
-        institution_key = ndb.Key(urlsafe=institution_key)
-        institution = institution_key.get()
+        user.state = 'inactive'
```
mayzabeel: Insert to do in the pull request to make tests for this class

backend/handlers/user_institutions_handler.py
raonismaneoto: Because actually we're handling with a collection of institutions.

PR #919, opened 2018-03-26
Fix remove of institution in config profile

tiagolimpe:
**Feature/Bug description:** When the user removes an institution the list don't update immediately.

**Solution:** Filtering the list to update the user institutions immediately after removal.

**TODO/FIXME:** n/a


mayzabeel:
The bug is resolved in https://github.com/eciis/web/pull/917 by @raonismaneoto

PR #921, opened 2018-03-26
Fix accept request to create institution

raonismaneoto:
**Feature/Bug description:**
The permissions were being added to the wrong user. They should be added to the sender but were being added to the super admin.

**Solution:**
I've called sender.add_permissions instead of user.add_permissions.

**TODO/FIXME:** n/a


backend/test/institution_request_handler_test.py
```diff
@@ -8,6 +8,7 @@
 from models.institution import Address
 from models.request_institution import RequestInstitution
 from handlers.institution_request_handler import InstitutionRequestHandler
+import permissions
```
mayzabeel: Refact this test class to use mock

PR #929, opened 2018-03-28
Fix create institution

MaianaB:
**Feature/Bug description:** When the inactive user requests the creation of the institution, it can create without responding all the attributes of the address, because the object "address" has only the attribute "country"

**Solution:** Added array that contains all the attributes that should be answered when the country is Brazil.

**TODO/FIXME:** n/a

screenshot from 2018-03-28 09-54-39

### DIFF
--- a/frontend/institution/configInstDirective.js
+++ b/frontend/institution/configInstDirective.js
@@ -436,7 +436,6 @@
             if (institutionKey) {
                 loadInstitution();
             } else {
-                console.log(">>>>>>>>>>>>>>. é novo");
                 configInstCtrl.isSubmission = true;
                 configInstCtrl.newInstitution.admin = {};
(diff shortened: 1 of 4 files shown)


frontend/institution/configInstDirective.js
```diff
@@ -360,12 +360,12 @@
 
         configInstCtrl.isValidAddress =  function isValidAddress(){       
             var valid = true;
-            var address = configInstCtrl.address;    
+            var address = configInstCtrl.address;
+            var fields = ['city', 'country', 'federal_state', 'neighbourhood', 'street', 'cep'];
```
ruanres: Em vez de validar, quando a instituição ainda não existe, é melhor criar um objeto Adress com estes campos vazios.

PR #930, opened 2018-03-28
Fix send invit user adm

Luiz-FS:
**Feature/Bug description:** The admin transfer page was allowing the admin to send an invitation to himself.

**Solution:** Block, both on the server and on the client, sending the admin transfer invitation to itself.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/invites/processInviteUserAdmController.js
+++ b/frontend/invites/processInviteUserAdmController.js
@@ -17,11 +17,7 @@
         processCtrl.accept = function accept() {
             InviteService.acceptInviteUserAdm(processCtrl.invite.key).then(function success() {
                 processCtrl.current_user.institutions_admin.push(processCtrl.invite.institution_key);
-                let institution = processCtrl.current_user.institutions.reduce(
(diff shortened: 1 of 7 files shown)


frontend/invites/processInviteUserAdmController.js
```diff
@@ -17,6 +17,11 @@
         processCtrl.accept = function accept() {
             InviteService.acceptInviteUserAdm(processCtrl.invite.key).then(function success() {
                 processCtrl.current_user.institutions_admin.push(processCtrl.invite.institution_key);
+                let institution = processCtrl.current_user.institutions.reduce(
```
mayzabeel: put this part in other function and document why it is necessary.

frontend/invites/processInviteUserAdmController.js
```diff
@@ -17,6 +17,11 @@
...
+                    (inst_found, institution) => (institution.key === processCtrl.invite.institution_key) ? institution : inst_found,
```
ruanres: usa cammelCase

PR #931, opened 2018-03-28
Improving the deleted_post notification

raonismaneoto:
**Feature/Bug description:**
Setting up an event listener that allows to update the post and remove it from the timeline when there is a deleted_post notification.
**Solution:**
I've created a event listener in timelineController that wraps the logic of delete the post from the timeline.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/service_messages.py
+++ b/backend/service_messages.py
@@ -55,7 +55,7 @@ def create_entity(entity_key):
     return json.dumps(entity)
 
 
-def send_message_notification(receiver_key, sender_key, entity_type, entity_key, current_institution=None):
+def send_message_notification(receiver_key, sender_key, entity_type, entity_key, current_institution=None, entity=None):
     """Method of send notification.
 
     Keyword arguments:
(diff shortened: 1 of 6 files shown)


frontend/post/post.js
```diff
@@ -53,3 +53,11 @@ Post.prototype.remove = function remove(userName) {
     this.last_modified_by = userName;
     this.last_modified_date = new Date().toISOString();
 };
+
+
+Post.prototype.hasActivity = function hasActivity() {
+    var hasNoComments = this.number_of_comments === 0;
```
mayzabeel: is better to understand hasComments and hasLikes, when hasNoComments = false is not good to read, the same to !hasNoComments

raonismaneoto: that's true

backend/service_messages.py
```diff
@@ -67,7 +67,7 @@ def send_message_notification(receiver_key, sender_key, entity_type, entity_key,
     current_institution -- the institution the user was logged in when the notification was sent
```
ruanres: adiciona a descrição desse novo parâmetro

PR #936, opened 2018-03-28
Fix remove member notification action

raonismaneoto:
**Feature/Bug description:**
Handle with current_institution in the backend became a problem to us in this situation. The user who the request came from wasn't a current_institution's member anymore, what make the server raise an exception and making the notification's action work with problem.
**Solution:**
I've changed the current_institution before make the request to the server to reload the user.

**TODO/FIXME:** n/a


mayzabeel:
:+1:

PR #953, opened 2018-04-03
Fix show message of empty posts in institution page

Luiz-FS:
**Feature/Bug description:** The message that the institution does not have publications was not appearing when a post that has no activity was deleted and the institution had no more post.

screenshot from 2018-04-03 10-01-22

**Solution:** I blocked on the server the sending of deleted posts and no activity and added a check in the frontend to verify that in the list of posts only contains deleted posts and no activity, without yes, shows the message, otherwise it does not show.

screenshot from 2018-04-03 10-01-50

**TODO/FIXME:** n/a

### DIFF
--- a/backend/handlers/institution_timeline_handler.py
+++ b/backend/handlers/institution_timeline_handler.py
@@ -41,9 +41,11 @@ class InstitutionTimelineHandler(BaseHandler):
             queryPosts)
 
         array = [post.make(self.request.host) for post in queryPosts]
+        visible_posts = [post for post in array
+                             if not Post.is_hidden(post)]
 
         data = {
-            'posts': array,
+            'posts': visible_posts,
             'next': more
         }
 
--- a/backend/handlers/institution_timeline_handler.py
+++ b/backend/handlers/institution_timeline_handler.py
@@ -40,8 +40,8 @@ class InstitutionTimelineHandler(BaseHandler):
             limit,
             queryPosts)
 
-        array = [post.make(self.request.host) for post in queryPosts]
-        visible_posts = [post for post in array
+        formated_posts = [post.make(self.request.host) for post in queryPosts]
+        visible_posts = [post for post in formated_posts
                              if not Post.is_hidden(post)]
 
         data = {
--- a/backend/test/institution_timeline_handler_test.py
+++ b/backend/test/institution_timeline_handler_test.py
@@ -106,65 +106,29 @@ class InstitutionTimelineHandlerTest(TestBaseHandler):
         """Test the institution_timeline_handler get method whit deleted post."""
         user = mocks.create_user()
         institution = mocks.create_institution()
-        #institution.address = mocks.create_address()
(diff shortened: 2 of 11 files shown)


backend/handlers/institution_timeline_handler.py
```diff
@@ -41,9 +41,11 @@ def get(self, user, url_string):
             queryPosts)
 
         array = [post.make(self.request.host) for post in queryPosts]
+        visible_posts = [post for post in array
```
mayzabeel: test

backend/handlers/institution_timeline_handler.py
```diff
@@ -41,9 +41,11 @@ def get(self, user, url_string):
             queryPosts)
 
         array = [post.make(self.request.host) for post in queryPosts]
```
ruanres: muda nome dessa variável para formated_posts ou algo similar

backend/handlers/institution_timeline_handler.py
```diff
@@ -40,10 +40,12 @@ def get(self, user, url_string):
             limit,
             queryPosts)
 
-        array = [post.make(self.request.host) for post in queryPosts]
+        formated_posts = [post.make(self.request.host) for post in queryPosts]
+        visible_posts = [post for post in formated_posts
```
raonismaneoto: Check if you can get the visible posts in the query.

backend/test/institution_timeline_handler_test.py
```diff
@@ -99,4 +99,36 @@ def test_get(self, verify_token):
...
+    def test_get_with_deleted_post(self, verify_token):
+        """Test the institution_timeline_handler get method whit deleted post."""
+        user = mocks.create_user()
+        institution = mocks.create_institution()
+        user.add_institution(institution.key)
+        institution.add_member(user)
+        
+        post = mocks.create_post(user.key, institution.key)
+        post.last_modified_by = user.key
+        ohter_post = mocks.create_post(user.key, institution.key)
```
MaianaB: other *

backend/test/institution_timeline_handler_test.py
```diff
@@ -99,4 +99,36 @@ def test_get(self, verify_token):
         self.assertEqual(post_last["text"], "At vero eos et accusamus et iusto",
                          "The text is not the expected one")
         self.assertEqual(post_last["state"], 'published',
-                         "The state of post should be published") '''
+                         "The state of post should be published")
+    
+    @patch('utils.verify_token', return_value=USER)
+    def test_get_with_deleted_post(self, verify_token):
+        """Test the institution_timeline_handler get method whit deleted post."""
```
raonismaneoto: whit?

backend/handlers/institution_timeline_handler.py
Luiz-FS: Have to do it this way because the is_hidden method uses the make of the post and not the post itself.

PR #959, opened 2018-04-04
Fix null event on event page

tiagolimpe:
**Feature/Bug description:** When the event page is opened the console displays the following error: cannot read property 'institution_key' of null.

**Solution:** Adding verification if the event is defined before the return of functions 'isInstitutionAdmin' and 'canChange'.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/event/eventDetailsDirective.js
+++ b/frontend/event/eventDetailsDirective.js
@@ -59,7 +59,8 @@
         };
 
         eventCtrl.canChange = function canChange(event) {
-            return eventCtrl.isEventAuthor(event) || isInstitutionAdmin(event);
+            if(event)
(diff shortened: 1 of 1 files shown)


frontend/event/eventDetailsDirective.js
```diff
@@ -59,7 +59,8 @@
         };
 
         eventCtrl.canChange = function canChange(event) {
-            return eventCtrl.isEventAuthor(event) || isInstitutionAdmin(event);
+            if(event)
+                return eventCtrl.isEventAuthor(event) || isInstitutionAdmin(event);
```
ruanres: faz 
return event && (eventCtrl.isEventAuthor(event) || isInstitutionAdmin(event)) ;

PR #965, opened 2018-04-05
Fix change institution profile color

Luiz-FS:
**Feature/Bug description:** When you change the cover color of an institution and enter the user edit information page and try to edit an institution's profile, the editing dialog does not open.

**Solution:** I changed how to clone the user when the color change dialog of the institution cover is opened because the cloning method previously used could not clone lists, and this caused bugs in the edition of user profiles because of this method transformant lists in objects which are not lists.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/home/homeController.js
+++ b/frontend/home/homeController.js
@@ -245,7 +245,7 @@
         };
 
         function loadProfile(){
-            colorPickerCtrl.newUser =  Utils.clone(colorPickerCtrl.user);
+            colorPickerCtrl.newUser =  _.cloneDeep(colorPickerCtrl.user);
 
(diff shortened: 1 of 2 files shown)


frontend/home/homeController.js
```diff
@@ -245,7 +245,7 @@
         };
 
         function loadProfile(){
-            colorPickerCtrl.newUser =  Utils.clone(colorPickerCtrl.user);
+            colorPickerCtrl.newUser =  _.cloneDeep(colorPickerCtrl.user);
```
mayzabeel: remove clone of Utils

PR #971, opened 2018-04-06
Fix goToDifferentInstitution function

raonismaneoto:
**Feature/Bug description:**
The function wasn't handling with the hierarchy

**Solution:**
I've adapted the notification's entity to wrap the remove_hierarchy information and used that to change to the right institution when the its value is true. 

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/home/homeController.js
+++ b/frontend/home/homeController.js
@@ -23,7 +23,6 @@
         homeCtrl.stateView = "";
 
         homeCtrl.user = AuthService.getCurrentUser();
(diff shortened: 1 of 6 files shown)


frontend/home/homeController.js
```diff
@@ -23,6 +23,7 @@
         homeCtrl.stateView = "";
 
         homeCtrl.user = AuthService.getCurrentUser();
+        console.log(homeCtrl.user);
```
mayzabeel: remove

PR #974, opened 2018-04-09
Create json patch observer recorder

Luiz-FS:
**Feature/Bug description:** When using the jsonpatch library to generate entity modification patches, this library generates an observer so that the changes that have been made to the entity can be verified, but this observer does not die when the patch is generated, it is still active causes any subsequent modification to the entity to be caught by the patch, generating garbage and possible bugs in the system since it will still be observing the entity when it should have already stopped. The current way we use to solve this problem is bad, because we only remove the garbage we previously forewarned from the patch, and we do not make the observer stop observing as it should be.

**Solution:** I have created a service (JsonPatchObserverRegister) to record all the observations created by jsonpatch, so that when finished the changes in the user and generate the patch, or leave the page, these generated observers are reset, eliminating any possible garbage that has been left. To observe the page change, I created an interceptor in the app.js (JsonPatchUnobserverInterceptor), which checks for any change of state and resets all observers that are registered in the service.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/auth/configProfileController.js
+++ b/frontend/auth/configProfileController.js
@@ -4,7 +4,7 @@
     var app = angular.module("app");
 
     app.controller("ConfigProfileController", function ConfigProfileController($state, InstitutionService,
-        CropImageService, AuthService, UserService, ImageService, $rootScope, $mdToast, $q, MessageService, $mdDialog, JsonPatchObserverRecorder) {
+        CropImageService, AuthService, UserService, ImageService, $rootScope, $mdToast, $q, MessageService, $mdDialog, JsonPatchObserverRecorderService) {
 
         var configProfileCtrl = this;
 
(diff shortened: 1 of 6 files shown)


frontend/utils/jsonPatchObserverRecorderService.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+(function() {
+    var app = angular.module('app');
+
+    app.service('JsonPatchObserverRecorderService', function() {
```
ruanres: The name is too long, could it be 'ObserverRecorderService'?

frontend/auth/configProfileController.js
```diff
@@ -74,18 +74,11 @@
             }
         };
 
-        function filter_patch(patch) {
-            return patch.filter((operation) => {
-                var atomicPath = operation.path.split('/')[1];
-                return CAN_EDIT.includes(atomicPath);
```
mayzabeel: Remove the const CAN_EDIT too

Luiz-FS: Done

PR #975, opened 2018-04-09
Refact the method get_deciis 

mayzabeel:
**Feature/Bug description:** The method to get the super-institution DECIIS, get the institution filtered by name. But if the name of institution change, the logic doesn't work anymore.

**Solution:** Get the super institution filtering by the flag trusted

**TODO/FIXME:** Change the DECIIS flag trusted to true on production environment.

### DIFF
--- a/backend/handlers/institution_followers_handler.py
+++ b/backend/handlers/institution_followers_handler.py
@@ -64,7 +64,7 @@ class InstitutionFollowersHandler(BaseHandler):
         Utils._assert(institution.name == 'Ministério da Saúde',
                       "The institution can not be unfollowed", NotAuthorizedException)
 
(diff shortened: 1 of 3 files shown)


backend/handlers/institution_followers_handler.py
```diff
@@ -64,7 +64,7 @@ def delete(self, user, url_string):
         Utils._assert(institution.name == 'Ministério da Saúde',
                       "The institution can not be unfollowed", NotAuthorizedException)
 
-        Utils._assert(institution.name == 'Departamento do Complexo Industrial e Inovação em Saúde',
+        Utils._assert(institution.trusted == True,
```
raonismaneoto: You can pass just institution.trusted to Utils._assert

PR #979, opened 2018-04-10
Fix requested invitations in invite institution's page

raonismaneoto:
**Feature/Bug description:**
It was missing the json_response decorator in the delete method and It should open the dialog when the md-list-item is clicked.

**Solution:**
I've called openPendingRequestsDialog when the md-list-item is clicked, removed the accept and reject button and their functions and tests.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/invites/invite_institution.html
+++ b/frontend/invites/invite_institution.html
@@ -90,7 +90,7 @@
             <div flex >
               <md-list class="md-dense" flex>
                 <md-list-item class="md-3-line" ng-repeat="request in inviteInstCtrl.sent_requests"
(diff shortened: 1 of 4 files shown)


frontend/invites/invite_institution.html
```diff
@@ -90,18 +90,14 @@ <h1>
             <div flex >
               <md-list class="md-dense" flex>
                 <md-list-item class="md-3-line" ng-repeat="request in inviteInstCtrl.sent_requests"
-                ng-click="inviteInstCtrl.showProcessingRequestDialog(request)">
```
mayzabeel: remove showProcessingRequestDialog

PR #986, opened 2018-04-11
Fix long url bigger than limit of text in post

tiagolimpe:
**Feature/Bug description:** If has a bigger URL in a post and this URL exceeds the limit of chars to display in the timeline the post shows the HTML tags and don't generate a hyperlink.

**Solution:** Creating an algorithm to find the HTML tags to show the hyperlink and verify if the end of tags is above of limit of the text, isolate the position of the first occurrence which exceeds the limit of text and replaces the limit of text to support the complete hyperlink.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/utils/utils.js
+++ b/frontend/utils/utils.js
@@ -64,15 +64,17 @@ var Utils = {
         return indexes;
     },
     limitString : function limitString(string, limit) {
-        var undefinedIndex = -1;
-        var endIndexesOfLast = this.getIndexesOf("</a>", string);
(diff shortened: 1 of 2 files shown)


frontend/utils/utils.js
```diff
@@ -51,10 +51,30 @@ var Utils = {
 
         config.url = config.url.replace(restApiRegex, restApiUrl + '/api/$1');
     },
+    getIndexesOf : function getIndexesOf(substring, string) {
```
mayzabeel: better define
 `indexes = []`
`if (substring.length != 0) `
`....`
`return indexes`

PR #988, opened 2018-04-12
Add ndb.transactional in add comment method of post

MaianaB:
**Feature/Bug description:** When multiple users commented on the same post at the same time competition error occurs.

**Solution:**  Added ndb.transaction in internal scope to try to resolve this.

**TODO/FIXME:**Create deslike_comment method and replace ndb.transactional to internal scopes.

### DIFF
--- a/backend/handlers/like_handler.py
+++ b/backend/handlers/like_handler.py
@@ -76,6 +76,10 @@ class LikeHandler(BaseHandler):
 
             enqueue_task('post-notification', params)
 
+    """TODO:  Create deslike_comment method and replace 
+            ndb.transactional to internal scopes.
+        @author: Maiana Brito 12/04/2018
+    """
     @json_response
     @login_required
     @ndb.transactional(xg=True)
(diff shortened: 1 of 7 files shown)


backend/models/post.py
```diff
@@ -253,6 +253,7 @@ def get_number_of_comment(self):
         """Get number of comments."""
         return len(self.comments)
 
+    @ndb.transactional(retries=10)
     def add_comment(self, comment):
         """Add a comment to the post."""
         self.comments[comment.id] = Utils.toJson(comment)
```
mayzabeel: is necessary get the entity before update

backend/handlers/like_handler.py
```diff
@@ -76,6 +76,10 @@ def post(self, user, post_key, comment_id=None, reply_id=None):
 
             enqueue_task('post-notification', params)
 
+    """TODO:  Create deslike_comment method and replace
```
raonismaneoto: dislike I guess

PR #999, opened 2018-04-16
Send notification when you reject the request to become a member of an institution.

MaianaB:
**Feature/Bug description:** When admin rejects the request to be a member of the institution, the user didn't receive any return. 

**Solution:** Send notification.

**TODO/FIXME:** n/a

screenshot from 2018-04-16 15-38-49

### DIFF
--- a/backend/handlers/request_handler.py
+++ b/backend/handlers/request_handler.py
@@ -77,15 +77,7 @@ class RequestHandler(BaseHandler):
 
         host = self.request.host
         request.send_response_email(host, "ACCEPT")
-
-        entity_type = 'ACCEPTED_LINK'
-        send_message_notification(
-            request.sender_key.urlsafe(),
-            user.key.urlsafe(),
-            entity_type,
-            institution.key.urlsafe(),
-            user.current_institution
-        )
+        request.send_response_notification(user, institution.key, 'ACCEPTED_LINK')
 
         self.response.write(json.dumps(makeUser(sender, self.request)))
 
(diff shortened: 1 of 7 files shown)


backend/handlers/request_handler.py
```diff
@@ -116,3 +115,14 @@ def delete(self, user, request_key):
 
         host = self.request.host
         request.send_response_email(host, "REJECT")
+
+        entity_type = 'REJECTED_LINK'
```
Luiz-FS: Move this logic to request user model

frontend/notification/notifications_page.html
```diff
@@ -25,7 +25,7 @@ <h1>
                             <img ng-src="{{notification.from.photo_url}}" class="md-avatar" style="width: 60px; height: 60px;"/>
                             <div class="md-list-item-text" layout="column">
                                 <p><b style="font-size: 18px;">{{ notificationCtrl.limitString(notification.from.name, 85) }}</b></p>
-                                <p><b style="font-size: 15px; color: #009688;">{{ notificationCtrl.limitString(notification.from.institution_name, 85) }}</b></p>
+                                <p><b style="font-size: 15px; color: #009688;">{{ notificationCtrl.limitString(notification.entity.institution_name, 85) }}</b></p>
```
Luiz-FS: Check this modification on other notifications

backend/models/request_user.py
```diff
@@ -101,6 +102,22 @@ def send_notification(self, current_institution):
...
 
+    """ 
+    The sender, in this case, is the user who is asking to be an institution's members.
+    The user is who answered the request.
+    The institution_key to which the sender requested to be member.
+    Entity_type can be REJECTED_LINK or ACCEPTED_LINK.
+    """
+    def send_response_notification(self, user, institution_key, entity_type):
+        """Method sends response notification to sender request."""
+        send_message_notification(
```
Luiz-FS: Check the possibility of generalizing this function. In the superclass invite, there already exists a 'send_response_notification' method, it always sends the invite key in the entity_key field, but could get to receive the entity_key as an optional parameter and if it isn't passed, use the invite key. This way it is easier to resolve the issue with the other invites that are also having problems with the notification.

PR #1009, opened 2018-04-18
Fix shared post notification

Luiz-FS:
**Feature/Bug description:** When sharing a post the notification that is sent is wrong because the institution that is sent in the notification is not the institution in which the original post was created.

screenshot from 2018-04-18 09-46-25

**Solution:** In the send_message_Notification method used in postNotificationsHandler I pass the key of the post that was shared when it exists, if there is no key to the original post.

screenshot from 2018-04-18 09-46-36

**TODO/FIXME:** n/a

### DIFF
--- a/backend/worker.py
+++ b/backend/worker.py
@@ -219,7 +219,7 @@ class PostNotificationHandler(BaseHandler):
                     subscriber,
                     sender_key,
                     entity_type,
-                    post_key,
+                    shared_entity_key or post_key,
                     current_institution
                 )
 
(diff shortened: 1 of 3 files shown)


backend/worker.py
```diff
@@ -219,7 +219,7 @@ def post(self):
                     subscriber,
                     sender_key,
                     entity_type,
-                    post_key,
+                    shared_entity_key or post_key,
```
mayzabeel: document the method say who is post_key and who is shared_entity_key

PR #1012, opened 2018-04-20
Fix show message if edit post is fail

Luiz-FS:
**Feature/Bug description:** When a user creates a post and opens the modal edit post and tries to edit it if the post already has action, this edition does not work and no message appears to the user informing that the operation did not work.

screenshot from 2018-04-20 10-12-19

**Solution:** I added an error message when the user can no longer edit the post and the post editing fails.

screenshot from 2018-04-20 10-12-27

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/post/postDirective.js
+++ b/frontend/post/postDirective.js
@@ -304,7 +304,7 @@
                             MessageService.showToast(response.data.msg);
                         });
                     }, function error() {
-                        MessageService.showToast("Esse post não pode ser editado pois já possue atividade");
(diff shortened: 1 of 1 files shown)


frontend/post/postDirective.js
```diff
@@ -303,6 +303,8 @@
                             $mdDialog.cancel();
                             MessageService.showToast(response.data.msg);
                         });
+                    }, function error() {
+                        MessageService.showToast("Esse post não pode ser editado pois já possue atividade");
```
mayzabeel: possui

PR #1014, opened 2018-04-20
Fix like post when deleted post

Luiz-FS:
**Feature/Bug description:** When a user creates a post and then deletes it, if another user who is viewing the post does not refresh the page, it still succeeds in like the post.

**Solution:** I added a check in the like_handler that checks if the post is deleted, if yes, does not let give like.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/handlers/like_handler.py
+++ b/backend/handlers/like_handler.py
@@ -8,6 +8,7 @@ from utils import json_response
 from handlers.base_handler import BaseHandler
 from models.post import Like
 from custom_exceptions.notAuthorizedException import NotAuthorizedException
+from custom_exceptions.entityException import EntityException
 from service_entities import enqueue_task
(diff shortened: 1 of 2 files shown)


backend/handlers/like_handler.py
```diff
@@ -48,6 +49,8 @@ def post(self, user, post_key, comment_id=None, reply_id=None):
         """Handle POST Requests."""
         """This method is only meant to give like in post, comment or reply and send notification."""
         post = ndb.Key(urlsafe=post_key).get()
+        Utils._assert(post.state == 'deleted',
```
raonismaneoto: make tests for it, thus  you guarantee that any changes in this assert will be seen.

Luiz-FS: Done.

PR #1018, opened 2018-04-20
Improving hierarchy requests

tiagolimpe:
**Feature/Bug description:** When an admin invites child institution the institution is not added to your children_institutions list. On another hand, the invite to parent institution add the institution invited to your parent institution. 

**Solution:** Removing the display of requests to who send the invite and displaying only to who receive the invite. Also, displaying the institution status on the institutional_links page.

screenshot from 2018-04-20 16-22-34

screenshot from 2018-04-20 16-22-17

**TODO/FIXME:** n/a


ruanres:
Atualizar os testes para verificar se as propriedades parent_institution e children_institutions das instituição que estão sendo linkadas

PR #1020, opened 2018-04-24
Fix concurrence problem in reply comment.

MaianaB:
**Feature/Bug description:** When many users reply the comment in the same moment, returns concurrence error.

**Solution:** Added ndb.transactional in the method that adds the reply.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/handlers/reply_comment_handler.py
+++ b/backend/handlers/reply_comment_handler.py
@@ -40,7 +40,6 @@ class ReplyCommentHandler(BaseHandler):
 
     @json_response
     @login_required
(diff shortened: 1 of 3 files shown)


backend/handlers/reply_comment_handler.py
```diff
@@ -55,7 +55,6 @@ def post(self, user, post_key, comment_id):
 
         reply = Comment.create(reply_data, user)
         post.reply_comment(reply, comment_id)
-        post.put()
```
mayzabeel: remove @ndb.transactional(xg=True) of method post

PR #1023, opened 2018-04-25
Fix remove permissions when a link between two institutions is undone

ruanres:
**Feature/Bug description:** When the link is undone, the admin that requested the unlink was losing permissions to some other institutions that were children of the unlinked institution, for example, in the following scenario:            **A -> B -> C -> D**
where, A, B, C and D are institutions and admA == AdmC and admB == admD.
So, when the link between A and B is undone, admA also lose permission to institution D. This occurs because all permissions from institutions hierarchically underneath B are retrieved and the method to check if the permissions from one of these institutions, should be removed, e.g. X, only verifies if admA is the same as admX, and does not consider the case in which admA can be admin of one institution hierarchically above X.

**Solution:** Now, the method to retrieve the permissions only gets the permission until it finds a child institution in which its admin is the same as the admin that requested the unlink 

**TODO/FIXME:** Update tests to check the case above and the ones in which there is administration transfer.


backend/models/institution.py
```diff
@@ -319,15 +319,18 @@ def is_active(self):
...
+    def get_hierarchy_admin_permissions(self, get_all=True, admin_key=None, permissions=None):
         """
         This method get all hierarchy admin permissions of
         institution. When result, returns a dict containing
         the admin permissions of all institutions belonging 
         to the child hierarchy.
 
         Arguments:
-        permissions(Optional) -- Dict of previous permissions added for add more permissons. 
+        get_all (Optional)-- If false, get all permissions from institutions
```
mayzabeel: and if true?

backend/worker.py
```diff
@@ -344,31 +347,39 @@ def post(self):
 class TransferAdminPermissionsHandler(BaseHandler):
     """Handler of transfer admin permissions."""
 
-    def add_permissions(self, user, permissions):
+    def add_permissions(self, user, institution):
         """
         This method adds new permissions in user with the permissions passed in parameter.
```
mayzabeel: update this comment

backend/worker.py
```diff
@@ -344,31 +347,39 @@ def post(self):
...
         for permission in permissions:
             if permission in user.permissions:
                 user.permissions[permission].update(permissions[permission])
             else:
                 user.permissions.update({permission: permissions[permission]})
     
-    def remove_permissions(self, user, permissions):
+    def remove_permissions(self, user, institution):
         """    
         This method removes the permissions of the user according to the permissions
```
mayzabeel: update comment

backend/models/institution.py
```diff
@@ -319,15 +319,18 @@ def is_active(self):
...
-    def get_all_hierarchy_admin_permissions(self, permissions=None):
+    def get_hierarchy_admin_permissions(self, get_all=True, admin_key=None, permissions=None):
         """
         This method get all hierarchy admin permissions of
         institution. When result, returns a dict containing
         the admin permissions of all institutions belonging 
         to the child hierarchy.
 
         Arguments:
-        permissions(Optional) -- Dict of previous permissions added for add more permissons.
```
raonismaneoto: I agree with @Luiz-FS there is some semantic error. get_all = True gotta mean: The method will get all the permissions.

backend/worker.py
raonismaneoto: permissions passed as a parameter? It's showing here that you're passing institution instead of the previous parameter, permissions.

backend/worker.py
raonismaneoto: This comment is strange, permissions is passed as parameter to where?

raonismaneoto: Test it

backend/worker.py
raonismaneoto: test it

backend/models/institution.py
```diff
@@ -341,9 +344,12 @@ def get_all_hierarchy_admin_permissions(self, permissions=None):
             else:
                 permissions.update({permission: {institution_key: True}})
 
-        for children in self.children_institutions:
-            children = children.get()
-            children.get_all_hierarchy_admin_permissions(permissions)
+        for child_key in self.children_institutions:
```
raonismaneoto: test

backend/models/institution.py
```diff
@@ -319,15 +319,18 @@ def is_active(self):
         """Check if is active."""
         return self.state == "active"
     
-    def get_all_hierarchy_admin_permissions(self, permissions=None):
+    def get_hierarchy_admin_permissions(self, get_all=True, admin_key=None, permissions=None):
```
raonismaneoto: test

PR #1024, opened 2018-04-25
Fix add permissions when accept instituion link

Luiz-FS:
**Feature/Bug description:** When a user agrees to link between two institutions, their permissions are not updated, requiring that the user reload the application for the permissions to be updated.

**Solution:** I created a method in requestProcessingController so that once the user accepts the link, their permissions are reloaded. And on the user receiving the notification that the invitation to create link was accepted, as soon as the invitation clicks on the notification, their permissions are reloaded.

**TODO/FIXME:** n/a


frontend/requests/requestProcessingController.js
```diff
@@ -25,11 +25,19 @@
                 MessageService.showToast("Solicitação aceita!");
                 requestController.request.status = 'accepted';
                 requestController.hideDialog();
+                refreshUser();
```
raonismaneoto: make tests to assert that the permissions are being added

Luiz-FS: Done.

PR #1038, opened 2018-05-02
Fix dislike concurrency

raonismaneoto:
**Feature/Bug description:**
The ndb.transaction was decorating all the delete method.
**Solution:**
I've moved the decorator to a more internal scope.
**TODO/FIXME:** n/a

### DIFF
--- a/backend/handlers/like_handler.py
+++ b/backend/handlers/like_handler.py
@@ -86,7 +86,6 @@ class LikeHandler(BaseHandler):
     """
     @json_response
     @login_required
-    @ndb.transactional(xg=True)
     def delete(self, user, post_key, comment_id=None, reply_id=None):
         """Handle DELETE Requests."""
         """This method is only meant to dislike in post."""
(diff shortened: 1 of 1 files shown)


backend/handlers/like_handler.py
```diff
@@ -105,10 +104,18 @@ def delete(self, user, post_key, comment_id=None, reply_id=None):
 
             Utils._assert(user.key.urlsafe() not in likes,
                       "User hasn't liked this comment.", LikeException)
-            likes.remove(user.key.urlsafe())
-            post.put()
+
+            @ndb.transactional(retries=10, xg=True)
+            def remove_like(likes, user, post):
+                likes.remove(user.key.urlsafe())
```
mayzabeel: you have to make a get again to guarantee that is the more actualized data

backend/handlers/like_handler.py
```diff
@@ -105,10 +104,18 @@ def delete(self, user, post_key, comment_id=None, reply_id=None):
...
+            remove_like(likes, user, post)
         else:
             Utils._assert(not user.is_liked_post(post.key),
                       "User hasn't liked this publication.", LikeException)
-            user.dislike_post(post.key)
-            post.dislike(user.key)
+
+            @ndb.transactional(retries=10, xg=True)
+            def remove_like(user, post):
+                user.dislike_post(post.key)
```
mayzabeel: you have to make a get again to guarantee that is the more actualized data

PR #1041, opened 2018-05-03
Fix remove institution

raonismaneoto:
**Feature/Bug description:**
It was missing a verification that checks if the link is confirmed when going down in the hierarchy in the removal process.
**Solution:**
I've added the check in the places that should have it and refactored some methods to work according to the new logic. 

**TODO/FIXME:** n/a


backend/models/institution.py
```diff
@@ -211,15 +211,17 @@ def handle_hierarchy_removal(self, remove_hierarchy, user):
         user -- the user who started the institution's removal process. So, it may not be the admin
         of self.
         """
+        returned_callback_method = False
```
ruanres: it should be initialized with None

ruanres: Improve the name and description of this method

backend/test/institution_handler_test.py
```diff
@@ -581,6 +581,226 @@ def test_remove_admin_permission(self, verify_token, enqueue_task):
             first_user, second_inst.key.urlsafe(), permissions.DEFAULT_ADMIN_PERMISSIONS))
         self.assertTrue(has_permissions(
             first_user, fourth_inst.key.urlsafe(), permissions.DEFAULT_ADMIN_PERMISSIONS))
+    
+    @patch('handlers.institution_handler.enqueue_task')
+    @patch('utils.verify_token', return_value={'email': 'user@example.com'})
+    def test_delete_hierarchy_with_a_one_direction_link(self, verify_token, enqueue_task):
```
ruanres: To each test, you should add a comment to illustrate the configuration of this hierarchy. For example, look at RemoveAdminPermissionsInInstitutionHierarchyTest.py, line 53

backend/worker.py
```diff
@@ -233,10 +235,12 @@ def post(self):
 
         @ndb.transactional(xg=True, retries=10)
         def apply_remove_operation(remove_hierarchy, institution, user):
-            institution.handle_hierarchy_removal(remove_hierarchy, user)
+            callback_method = institution.handle_hierarchy_removal(remove_hierarchy, user)
```
Luiz-FS: This method return does not seem to be a callback, would it be better to leave that name just 'returned_method'.

PR #1042, opened 2018-05-04
Fix remove admin permissions handler

raonismaneoto:
**Feature/Bug description:**
The method wasn't going up in the hierarchy removing the permissions.
**Solution:**
I've got the parent_admins and iterate over them removing the permissions when it should be removed.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/test/remove_admin_permissions_hierarchy_test.py
+++ b/backend/test/remove_admin_permissions_hierarchy_test.py
@@ -92,7 +92,7 @@ class RemoveAdminPermissionsInInstitutionHierarchyTest(TestBaseHandler):
         first_user.follows.append(first_inst.key)
         first_user.put()
 
-        params = '?institution_key=%s&&user=%s' % (second_inst.key.urlsafe(), first_user.key.urlsafe())
+        params = '?institution_key=%s&&parent_key=%s' % (second_inst.key.urlsafe(), first_inst.key.urlsafe())
         self.testapp.post(
             REMOVE_ADMIN_PERMISSIONS_URI + params
         )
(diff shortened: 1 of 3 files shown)


backend/worker.py
```diff
@@ -353,17 +353,27 @@ def post(self):
...
-            )
-            self.removeAdminPermissions(user, permissions)
-        apply_remove_operation(user, institution, is_not_admin)
+        def apply_remove_operation(parent_admin, institution, should_remove):
+            """This method is responsible for getting the permissions involved
+            in the link and go up in the hierarchy removing the permissions from
+            the admins that have to lose it, based in a condition that checks if 
+            the current_admin is different of the current_admin.  
+            """
+            admins = get_all_parent_admins(parent_institution, [])
```
mayzabeel: parent_admins

backend/test/remove_admin_permissions_hierarchy_test.py
```diff
@@ -107,3 +107,260 @@ def test_post(self):
...
+
+        first_user.institutions_admin.append(first_inst.key)
+        second_user.institutions_admin.append(second_inst.key)
+        second_user.institutions_admin.append(third_inst.key)
+        third_user.institutions_admin.append(fourth_inst.key)
+
+        # Hierarchy
+        #   first_inst -> second_inst -> third_inst -> fourth_inst
+        #   (first_user)  (second_user)   (second_user)  (third_user)
+        second_inst.parent_institution = first_inst.key
```
mayzabeel: try to change the administrator of fourth_inst to first_user and check if the permissions still ok

backend/worker.py
```diff
@@ -353,17 +353,31 @@ def post(self):
...
-            permissions = filter_permissions_to_remove(
-                user, permissions, institution.key, should_remove
-            )
-            self.removeAdminPermissions(user, permissions)
-        apply_remove_operation(user, institution, is_not_admin)
+        def apply_remove_operation(parent_admin, institution, should_remove):
+            """This method is responsible for getting the permissions involved
+            in the link and go up in the hierarchy removing the permissions from
+            the admins that have to lose it, based in a condition that checks if 
+            the current_admin is different of the current_admin.
```
ruanres: current_admin is different of the current_admin ?

PR #1044, opened 2018-05-04
Fix get_all_parent_admins

raonismaneoto:
**Feature/Bug description:**
This method wasn't checking if the link was confirmed.
**Solution:**
Check if the link is confirmed before call the method again to the parent_institution.
**TODO/FIXME:** n/a

### DIFF
--- a/backend/test/worker_aux_methods_test.py
+++ b/backend/test/worker_aux_methods_test.py
@@ -9,15 +9,7 @@ from worker import get_all_parent_admins
 
 class WorkerAuxMethodsTest(TestBaseHandler):
     """Test Remove admin permission in institution hierarchy."""
-
-    @classmethod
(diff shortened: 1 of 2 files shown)


backend/test/worker_aux_methods_test.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class WorkerAuxMethodsTest(TestBaseHandler):
+    """Test Remove admin permission in institution hierarchy."""
+
+    @classmethod
+    def setUp(cls):
+        """Provide the base for the tests."""
+        super(WorkerAuxMethodsTest, cls).setUp()
+        app = cls.webapp2.WSGIApplication(
```
Luiz-FS: If you are not testing handlers you do not need this initialization.

PR #1046, opened 2018-05-08
Forbid hierarchical loop

ruanres:
**Feature/Bug description:** It was possible to create a hierarchical loop. For example, a hierarchically higher institution could invite one of its subordinated institutions to be its parent.

**Solution:** It was created a verification, at the handlers responsible for the creation of parent and child request, to identify and forbid the creation possible hierarchical loop requests.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/handlers/institution_children_request_collection_handler.py
+++ b/backend/handlers/institution_children_request_collection_handler.py
@@ -9,6 +9,7 @@ from utils import Utils
 from custom_exceptions.entityException import EntityException
 from custom_exceptions.notAuthorizedException import NotAuthorizedException
 from . import BaseHandler
+from models.institution import Institution
 from models.factory_invites import InviteFactory
 from models.request_institution_children import RequestInstitutionChildren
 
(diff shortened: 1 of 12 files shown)


backend/handlers/institution_parent_request_collection_handler.py
```diff
@@ -54,12 +55,23 @@ def post(self, user, institution_key):
             EntityException
         )
 
+        child_key = data.get('institution_key')
+        child_key = ndb.Key(urlsafe=child_key)
+        requested_inst_key = data.get('institution_requested_key')
+        requested_inst_key = ndb.Key(urlsafe=requested_inst_key)
+
+        Utils._assert(
+            Institution.has_connection_between(requested_inst_key, child_key),
```
mayzabeel: here you pass parent and child, but the order in the function is the opposite
`def has_connection_between(possible_child_key, possible_parent_key):`

backend/handlers/institution_children_request_collection_handler.py
```diff
@@ -52,11 +53,22 @@ def post(self, user, institution_key):
             EntityException
         )
 
+        parent_key = data.get('institution_key')
+        parent_key = ndb.Key(urlsafe=parent_key)
+        requested_inst_key = data.get('institution_requested_key')
+        requested_inst_key = ndb.Key(urlsafe=requested_inst_key)
+        
+        Utils._assert(
+            Institution.has_connection_between(parent_key, requested_inst_key),
```
mayzabeel: here you pass parent and child, but the order in the function is the opposite
def has_connection_between(possible_child_key, possible_parent_key):

backend/handlers/institution_parent_request_collection_handler.py
ruanres: As this is the handler responsible for request a parent, it will create a hierarchical loop if the requested institution ( (requested_inst_key)) is a child of the institution that created the request (child_key). Besides that, the method has_connection_between makes a recursion ascending the hierarchy. That's why it is necessary to call this method with the requested_inst_key as the possible_child_key and the child_key as the possible_parent_key.

backend/handlers/institution_children_request_collection_handler.py
ruanres: Here, the logic is similar to what I explained for the institution_parent_request_collection_handler.

PR #1057, opened 2018-05-14
Fix request child institution

ruanres:
**Feature/Bug description:** Previously, when a already rejected child request and tries to send the same invite again, the request is duplicated in the list. Besides that, the requested institution was added again as a children institution.

inite2

**Solution:** On client, I added a verification to the method addChildInst, to check if that institution is already a child, if so, it is not added to the list. On server, I created a method to add a child, at Institution model, to prevent the same institution be added twice to the children institutions.

invite

**TODO/FIXME:** n/a

### DIFF
--- a/backend/models/institution.py
+++ b/backend/models/institution.py
@@ -136,7 +136,6 @@ class Institution(ndb.Model):
     def create_parent_connection(self, invite):
         """Make connections between parent and daughter institution."""
         self.add_child(invite.institution_key)
-        self.put()
 
         institution_children = invite.institution_key.get()
         institution_children.parent_institution = self.key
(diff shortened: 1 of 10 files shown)


backend/models/institution.py
```diff
@@ -126,10 +126,16 @@ def addInvite(self, invite):
...
+        """Add a new child to children_institutions."""
+        if child_key not in self.children_institutions:
+            self.children_institutions.append(child_key)
+            self.put()
+
     @ndb.transactional(xg=True)
     def create_parent_connection(self, invite):
         """Make connections between parent and daughter institution."""
-        self.children_institutions.append(invite.institution_key)
+        self.add_child(invite.institution_key)
```
mayzabeel: is unnecessary the put after because the method add_child has a put

PR #1060, opened 2018-05-14
Refact models modules imports

Luiz-FS:
**Feature/Bug description:**  The model's imports made in the backend are extensive and difficult to read.
Example:

```python
from models.invite_institution_parent import InviteInstitutionParent
```

**Solution:** I set up the __init__.py file so that it recognizes all packet models by simplifying imports and making reading easier.
Example:

```python
from models import InviteInstitutionParent
```

**Refactored models:**

* Invite
* InviteInstitution
* InviteInstitutionChildren
* InviteInstitutionParent
* InviteUser
* Request
* RequestUser
* RequestInstituionParent
* RequestInstituionChildren

**TODO/FIXME:** n/a


backend/models/__init__.py
```diff
@@ -1,13 +1,30 @@
...
+from .invite_institution_parent import *
+from .invite_user import *
+from .request import *
+from .request_user import *
+from .request_institution_parent import *
+from .request_institution_children import *
 
 __all__ = []
 
 __all__ += user.__all__
```
ruanres: Try to user a for loop to collect all the imports

PR #1073, opened 2018-05-17
Fix addChildInst function

raonismaneoto:
**Feature/Bug description:**
When the two or more children institution's invite were created they're not appearing in the view because the function that checks if they should be added to the institution.children_institutions or not was checking if there was an invite with the same key as the current invite and the invites didn't have any key, once they hasn't gone to the backend, so, the function wasn't allowing the addition because undefined === undefined was returning true, what made it realize that there was an invite with the same key. 
**Solution:**
I've checked if the invites have the key before checking if it was equals to the current invite's key.
**TODO/FIXME:** n/a


ruanres:
You better improve this bug description.

PR #1083, opened 2018-05-21
Refact util modules imports

Luiz-FS:
**Feature/Bug description:** The util modules imports made in the backend are extensive and difficult to read.
Example:

```python
from util.login_service import  login_required
```

**Solution:** I set up the __init__.py file so that it recognizes all packet models by simplifying imports and making reading easier.
Example:

```python
from util import  login_required
```
**Refactored util modules:**

* login_service
* strings_pt_br
* json_patch

**TODO/FIXME:** n/a


backend/models/user.py
```diff
@@ -3,8 +3,8 @@
 from google.appengine.ext import ndb
 from custom_exceptions.fieldException import FieldException
 from custom_exceptions.notAuthorizedException import NotAuthorizedException
-from util.provider_institutions import get_deciis
-from util.provider_institutions import get_health_ministry
+#from . import get_deciis
```
mayzabeel: ?? remove

PR #1085, opened 2018-05-21
Fix duplicated institution children

tiagolimpe:
**Feature/Bug description:** When the institution A requests to be the parent of institution B, institution B reject this request and B requests to be a child of A, the institution B appears duplicated in the institution A.

**Solution:** Verifying if the institutions have been connected before adding the link.

**TODO/FIXME:** n/a


backend/handlers/institution_parent_request_handler.py
```diff
@@ -33,11 +33,13 @@ def put(self, user, request_key):
         request.put()
 
         parent_institution = request.institution_requested_key.get()
-        parent_institution.children_institutions.append(request.institution_key)
-        parent_institution.put()
+
+        if request.institution_key not in parent_institution.children_institutions:
```
raonismaneoto: institution.py's add_child method verifies this condition and add the child. Use it

backend/handlers/institution_children_request_handler.py
```diff
@@ -54,6 +54,12 @@ def delete(self, user, request_key):
         request.change_status('rejected')
         request.put()
 
+        institution_children = request.institution_requested_key.get()
```
mayzabeel: make tests of all modifications

backend/handlers/institution_parent_request_handler.py
```diff
@@ -56,5 +58,11 @@ def delete(self, user, request_key):
         request.change_status('rejected')
         request.put()
 
+        parent_institution = request.institution_requested_key.get()
+
+        if request.institution_key in parent_institution.children_institutions:
```
raonismaneoto: I suggest you to create a remove_child method, the opposite of add_child, to keep pattern.

backend/handlers/institution_children_request_handler.py
```diff
@@ -54,6 +53,12 @@ def delete(self, user, request_key):
...
+
+        if institution_children.parent_institution == request.institution_key:
+            institution_children.parent_institution = None
```
ruanres: there is a method for that

backend/handlers/institution_parent_request_handler.py
```diff
@@ -56,5 +55,8 @@ def delete(self, user, request_key):
         request.change_status('rejected')
         request.put()
```
ruanres: this method already do a put

PR #1090, opened 2018-05-23
Fix create survey post

Luiz-FS:
**Feature/Bug description:** When creating a survey type post and trying to move to another page, the dialog indicating that unsaved changes appear, even after the post has already been created.

**Solution:** I created a function in the submitFormListener service that allows a given object to stop being observed, and I used it in the surveyController to stop observing the post created when it is posted.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/utils/submitFormListenerService.js
+++ b/frontend/utils/submitFormListenerService.js
@@ -43,8 +43,11 @@
 
         service.unobserve = function unobserve(obj){
             let transitionListener = listeners[obj];
-            transitionListener();
-            delete listeners[obj];
+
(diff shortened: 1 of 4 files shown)


frontend/utils/submitFormListenerService.js
```diff
@@ -25,16 +27,27 @@
                             'Deseja sair sem salvar as alterações?');
                     
                     promisse.then(function success() {
-                        transitionListener();
+                        service.unobserve(obj);
```
mayzabeel: update tests

PR #1093, opened 2018-05-24
Add button to analyse requests

Luiz-FS:
**Feature/Bug description:** The md-list of accepting institution and memebro requests did not make explicit that it could be clicked.

screenshot from 2018-05-24 09-45-55
screenshot from 2018-05-24 09-46-16

**Solution:** I created a button in the md-list so that the user when viewing the request knows that he can click on it.

screenshot from 2018-05-24 09-40-28
screenshot from 2018-05-24 09-41-03

**TODO/FIXME:** n/a


frontend/institution/management_members.html
```diff
@@ -250,6 +250,11 @@ <h3>{{ request.sender_name}}</h3>
               <h4>Cargo atual: {{ request.office}}</h4>
               <p>Email institucional: {{ request.institutional_email}}</p>
             </div>
+            <div layout="row" layout-align="end center" flex hide show-gt-xs>
+              <p class="status-invite-adm" style="line-height: 21px; padding: 5px;" md-colors="{color: 'teal-500'}">
```
ruanres: It is better to create a new ccs class for this button with the status-invite-adm properties and the props defined at the style tag

PR #1095, opened 2018-05-24
Fix accept member dialog

MaianaB:
**Feature/Bug description:** In the case of emails or names being large, they are off the card.

**Solution:** Defined different labels according to display size.

**TODO/FIXME:** n/a

screenshot from 2018-05-24 14-53-04

screenshot from 2018-05-24 14-54-59


frontend/requests/request_processing.html
```diff
@@ -26,8 +26,18 @@ <h4 style="margin: 0;">Confirmar Vínculo</h4>
                   alt="{{ requestCtrl.children.name }}"></img>
               </div>
               <div layout="column" layout-align="center start">
-                <b>{{ requestCtrl.children.name || requestCtrl.children.sender_name}}</b>
-                <span>{{ requestCtrl.children.institutional_email }}</span>
+                <b hide-gt-xs> {{ requestCtrl.limitString( requestCtrl.children.name ||
```
ruanres: Instead of call limitString, it would be better create a function to get the children name, that receives max size of the string and returns the formatted name

frontend/requests/request_processing.html
```diff
@@ -26,8 +26,18 @@ <h4 style="margin: 0;">Confirmar Vínculo</h4>
...
+                    requestCtrl.children.sender_name, 25)}}
+                </b>
+                <b hide-xs>{{ requestCtrl.limitString( requestCtrl.children.name ||
+                    requestCtrl.children.sender_name, 60)}}
+                </b>
+                <span hide-gt-xs>
+                  {{ requestCtrl.limitString(requestCtrl.children.institutional_email+'....', 28) }}
```
ruanres: same idea

Luiz-FS: Remove '....'

MaianaB:
Closed, problems with merge
#1107

MaianaB:
Problems with merge.
#1107

PR #1096, opened 2018-05-24
Refact institution handler's post method

raonismaneoto:
**Feature/Bug description:**
The method shouldn't be a post once it wasn't creating any new entity, it was only updating a institution's stub to change its state to active and update the data.
**Solution:**
The method became a put, what makes sense because it updates the entity.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/test/institution_handler_test.py
+++ b/backend/test/institution_handler_test.py
@@ -144,11 +144,11 @@ class InstitutionHandlerTest(TestBaseHandler):
             "Error! User is not allowed to edit institution")
 
     @patch('util.login_service.verify_token', return_value={'email': 'other_user@example.com'})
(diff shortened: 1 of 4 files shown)


backend/test/institution_handler_test.py
```diff
@@ -144,11 +144,11 @@ def test_patch(self, verify_token):
             "Error! User is not allowed to edit institution")
 
     @patch('util.login_service.verify_token', return_value={'email': 'other_user@example.com'})
-    def test_post(self, verify_token):
+    def test_put(self, verify_token):
         """Test the post_handler's post method."""
```
Luiz-FS: Fix docstring

PR #1102, opened 2018-05-25
Improving event_handler's get method

raonismaneoto:
**Feature/Bug description:**
The received param was out of pattern.
**Solution:**
I've renamed the param to keep it in pattern.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/handlers/event_handler.py
+++ b/backend/handlers/event_handler.py
@@ -18,9 +18,9 @@ class EventHandler(BaseHandler):
 
     @json_response
     @login_required
-    def get(self, user, url_string):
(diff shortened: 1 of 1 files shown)


backend/handlers/event_handler.py
```diff
@@ -18,9 +18,9 @@ class EventHandler(BaseHandler):
 
     @json_response
     @login_required
-    def get(self, user, url_string):
+    def get(self, user, event_urlsafe):
```
mayzabeel: improve documentation. Get specific event if it is not deleted

PR #1103, opened 2018-05-25
Fix send invite hiearchy in frontend

Luiz-FS:
**Feature/Bug description:** When sending a user invitation the invite_body attribute was being duplicated because of a modification made in sendInvite in the inviteService to remove inconsistency between the sending of user and hierarchy invitations.

**Solution:** Return the sendInvite method from the inviteService to the old logic and only adapt the sending of hierarchical invitations while maintaining consistency with the user invitations.

**TODO/FIXME:** n/a


frontend/invites/inviteInstHierarchieController.js
```diff
@@ -84,7 +84,8 @@
         inviteInstHierCtrl.sendInstInvite = function sendInstInvite(invite) {
             inviteInstHierCtrl.isLoadingSubmission = true;
             var deferred = $q.defer();
-            var promise = InviteService.sendInvite(invite);
+            var invite_body = {invite_body: invite}
```
raonismaneoto: rename this variable to invite, I think it's better

frontend/test/specs/inviteServiceSpec.js
```diff
@@ -35,7 +35,7 @@
             var result;
             service.sendInvite(inviteUser).then(function(data){
                 result = data;
-                body['data'] = {invite_body:inviteUser};
+                body['data'] = inviteUser;
```
mayzabeel: var promise = InviteService.sendInvite({invite_body: invite});

The inviteUser is not compatible with the real object received by service or is?

Luiz-FS: Done.

PR #1108, opened 2018-05-25
Improving post of institution_children_request_collection_handler

tiagolimpe:
**Feature/Bug description:** The parent_key is getting from body. But it is unnecessary because the parent_key is equal to institution_key, and the parameter provide the institution_key.

**Solution:** ...

**TODO/FIXME:** n/a

### DIFF
--- a/backend/handlers/institution_children_request_collection_handler.py
+++ b/backend/handlers/institution_children_request_collection_handler.py
@@ -35,7 +35,8 @@ class InstitutionChildrenRequestCollectionHandler(BaseHandler):
     @login_required
     @json_response
     def post(self, user, institution_urlsafe):
-        """Handler of post requests."""
+        """Handler of post requests. This method is called when an
+        institution requests to be parent of other institution."""
         user.check_permission(
             'send_link_inst_request',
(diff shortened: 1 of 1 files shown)


backend/handlers/institution_children_request_collection_handler.py
```diff
@@ -53,8 +53,7 @@ def post(self, user, institution_key):
             EntityException
         )
 
-        parent_key = data.get('institution_key')
-        parent_key = ndb.Key(urlsafe=parent_key)
+        parent_key = ndb.Key(urlsafe=institution_key)
```
raonismaneoto: rename institution_key to institution_urlsafe, it's our new pattern

backend/handlers/institution_children_request_collection_handler.py
```diff
@@ -34,12 +34,12 @@ def get(self, user, institution_key):
 
     @login_required
     @json_response
-    def post(self, user, institution_key):
+    def post(self, user, institution_urlsafe):
```
mayzabeel: improve the documentation

PR #1110, opened 2018-05-25
Improving the post method of institution_request_collection_handler

tiagolimpe:
**Feature/Bug description:** The parameter "institution_key" is not used but the function require 3 arguments because login_required need this.

**Solution:** Change the name to a generic name.

**TODO/FIXME:** There are other methods in backend with not used arguments. Change the name of this args to a generic name.


mayzabeel:
Closed because the name is more intuitive to understand

PR #1130, opened 2018-06-04
Fix send invite user frontend

Luiz-FS:
**Feature/Bug description:** When send a user invitation, an error occurs on the console stating that cannot read property invites of undefined.

**Solution:** Adjust the response of the InviteService.sendInviteUser method so that it no longer uses response.data.invites but response.invites.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/institution/managementMembersController.js
+++ b/frontend/institution/managementMembersController.js
@@ -122,7 +122,7 @@
                     MessageService.showToast(response.msg);
                 }, function error(response) {
                     manageMemberCtrl.isLoadingInvite = false;
-                    MessageService.showToast(response.data.msg);
+                    MessageService.showToast(response.msg);
                 });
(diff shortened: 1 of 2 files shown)


frontend/institution/managementMembersController.js
```diff
@@ -108,21 +108,21 @@
             var requestBody = {
                 invite_body: invite,
                 emails: emails
-            }
+            };
             
             if (manageMemberCtrl.isValidAllEmails(emails) && manageMemberCtrl.isUserInviteValid(invite)) {
                 manageMemberCtrl.isLoadingInvite = true;
                 var promise = InviteService.sendInviteUser(requestBody);
                 promise.then(function success(response) {
```
ruanres: Rename this parameter

PR #1134, opened 2018-06-05
Fix vote survey

MaianaB:
**Feature/Bug description:** If there is the survey and search survey, and any user votes for the original search, the search share options disappear. As the type_survey variable of the search is changed to false, the function that ranks if the post is the search does not recognize the post as such.

**Solution:** Put return in the request when voting the survey, and update the options of the survey.

**TODO/FIXME:** n/a


ruanres:
You should use the word survey instead of search, in the problem description.

PR #1142, opened 2018-06-07
Fix loop when create link between institutions

Luiz-FS:
**Feature/Bug description:** When an institution A sends a request to institution B to be daughter and B rejects, A passes an unconfirmed link to B. If A sends a request for B to be a parent, and B confirms an infinite loop error is generated in the worker.

**Solution:** Modify the verify_connection method on the institution template so that it only verifies whether one institution is the parent or child of other of the other and not both checks at the same time.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/models/institution.py
+++ b/backend/models/institution.py
@@ -369,7 +369,7 @@ class Institution(ndb.Model):
             adm_is_child_adm = child.admin == admin_key
             get_next_permissions = get_all or not adm_is_child_adm
             #The child permissions should only be added if the link is confirmed.
-            link_confirmed = child.verify_connection(self)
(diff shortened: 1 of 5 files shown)


backend/models/institution.py
```diff
@@ -408,14 +412,24 @@ def create_notification_message(self, user_key, current_institution_key,
             sender_institution_key= sender_institution_key or current_institution_key
         )
 
-    def verify_connection(self, institution_to_verify):
+    def verify_connection(self, institution_to_verify, verification_type='OTHERWISE'):
```
mayzabeel: improve tests

PR #1150, opened 2018-06-08
Fix edit post.

MaianaB:
**Feature/Bug description:** The user can edit post that institution was deleted or the post was deleted.

**Solution:** Create method that verify if user can edit post.

**TODO/FIXME:** n/a

screenshot from 2018-06-11 15-08-00
screenshot from 2018-06-11 15-13-50

### DIFF
--- a/backend/handlers/post_handler.py
+++ b/backend/handlers/post_handler.py
@@ -83,7 +83,7 @@ class PostHandler(BaseHandler):
         post = ndb.Key(urlsafe=post_urlsafe).get()
 
         Utils._assert(not post.can_edit(),
-                        "The user can not update this post",
+                        "This post cannot be updated",
                         NotAuthorizedException)
 
         """Apply patch."""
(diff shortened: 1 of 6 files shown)


backend/handlers/post_handler.py
```diff
@@ -90,14 +78,10 @@ def patch(self, user, post_urlsafe):
 
         post = ndb.Key(urlsafe=post_urlsafe).get()
 
-        Utils._assert(post.has_activity(),
+        Utils._assert(not post.can_edit(user),
```
mayzabeel: is better separate, because one thing is "the post cannot be edited" and another thing is "you don't have permission to edit this post"

backend/handlers/post_handler.py
```diff
@@ -88,16 +76,16 @@ def patch(self, user, post_urlsafe):
...
         data = self.request.body
 
+        user.check_permission("edit_post",	
+                                "User is not allowed to edit this post",	
+                               post_urlsafe)
+
         post = ndb.Key(urlsafe=post_urlsafe).get()
 
-        Utils._assert(post.has_activity(),
+        Utils._assert(not post.can_edit(),
```
mayzabeel: change to "This post cannot be updated"

mayzabeel: and add in message service too

PR #1155, opened 2018-06-11
Fix delete user account

MaianaB:
**Feature/Bug description:** If you have more than one institution deleting your account, some institutions are not removed because of how it is iterating over user institutions.
For example, the user has the following array [InstA, InstB]. For in its first step remove InstA, the second step would remove InstB, however, InstB is at index 0 of the array.

**Solution:** Iterate by user's institutions at the end.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/test/user_handler_test.py
+++ b/backend/test/user_handler_test.py
@@ -39,11 +39,21 @@ class UserHandlerTest(TestBaseHandler):
         cls.institution.add_member(cls.user)
         cls.institution.add_member(cls.other_user)
         cls.institution.change_state('active')
+        # create other institution
+        cls.other_institution = mocks.create_institution()
+        cls.other_institution.admin = cls.other_user.key
+        cls.other_institution.follow(cls.user.key)
+        cls.other_institution.follow(cls.other_user.key)
+        cls.other_institution.add_member(cls.user)
+        cls.other_institution.add_member(cls.other_user)
+        cls.other_institution.change_state('active')
         # update user
         cls.user.institutions.append(cls.institution.key)
         cls.user.add_institution_admin(cls.institution.key)
         # update other_user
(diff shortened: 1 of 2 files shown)


backend/test/user_handler_test.py
```diff
@@ -91,9 +101,9 @@ def test_delete(self, verify_token):
         verify_token._mock_return_value = {'email': self.other_user.email[0]}
         # check the user properties before delete it
         self.assertEquals(self.other_user.state, "active", "The user state should be 'active'")
-        self.assertEquals(self.other_user.institutions, [self.institution.key], 
+        self.assertEquals(self.other_user.institutions, [self.other_institution.key, self.institution.key],
```
mayzabeel: check if the user is removed from institutions too. This test only check the institutions of user is updated but no check if the user is removed from member's institution

backend/test/user_handler_test.py
```diff
@@ -111,17 +121,23 @@ def test_delete(self, verify_token):
...
+        self.assertTrue(self.other_user.key not in self.other_institution.members, 
+            "The user key should not be in institution members") 
+        self.assertTrue(self.other_user.key in self.other_institution.followers, 
+            "The user key should be in institution followers") 
 
         # assert user has no longer institutions and permissions
         self.assertEquals(self.other_user.state, "inactive", "The user state should be 'inactive'")
         self.assertEquals(self.other_user.institutions, [], "User institutions should be empty")
-        self.assertEquals(self.other_user.follows, [self.institution.key], "User institutions should not be empty")
+        self.assertEquals(self.other_user.follows, [self.other_institution.key, self.institution.key], "User institutions should not be empty")
```
mayzabeel: Institutions followed by user should not be empty

PR #1157, opened 2018-06-11
Fix remove user account

Luiz-FS:
**Feature/Bug description:** When a user deletes the account, the list of institutions that he or she follows is deleted.

**Solution:** Do not remove the list of institutions that the user follows when he deletes the account.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/test/user_handler_test.py
+++ b/backend/test/user_handler_test.py
@@ -115,13 +115,13 @@ class UserHandlerTest(TestBaseHandler):
         # assert institution has no longer the deleted user
         self.assertTrue(self.other_user.key not in self.institution.members, 
             "The user key should not be in institution members") 
-        self.assertTrue(self.other_user.key not in self.institution.followers, 
-            "The user key should not be in institution followers") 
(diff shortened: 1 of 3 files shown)


backend/test/user_handler_test.py
```diff
@@ -115,13 +115,13 @@ def test_delete(self, verify_token):
...
-        self.assertTrue(self.other_user.key not in self.institution.followers, 
-            "The user key should not be in institution followers") 
+        self.assertTrue(self.other_user.key in self.institution.followers, 
+            "The user key should be in institution followers") 
 
         # assert user has no longer institutions and permissions
         self.assertEquals(self.other_user.state, "inactive", "The user state should be 'inactive'")
         self.assertEquals(self.other_user.institutions, [], "User institutions should be empty")
-        self.assertEquals(self.other_user.follows, [], "User institutions should be empty")
+        self.assertEquals(self.other_user.follows, [self.institution.key], "User institutions should not be empty")
```
mayzabeel: check if the user remains in institution followers too

PR #1158, opened 2018-06-11
Standardize http services

raonismaneoto:
**Feature/Bug description:**
Some of the frontend's requests weren't using HttpService, what makes the code dirty.
It was like:
```javascript
$http.method(url, body).then(function success(response) {
     deferred.resolve(response);
}, function error(response) {
    deferred.reject(response);
});
return deferred.promise
```
**Solution:**
I've changed these requests using the HttpService.
Now, It is like:
```javascript
return HttpService.method(url, body);
```

With this approach we gotta be careful, mainly in the controllers, because the response returned by HttpService is the original response.data, so, we only have to take, as the request's response, the response and not response.data anymore.

for example:
previously the getPost function was like that:
```javascript
function getPost(postKey) {
  InviteService.getPost(postKey).then(function success(response) {
     ctrl.post = response.data;
  });
}
```

now it is like:
```javascript
function getPost(postKey) {
  InviteService.getPost(postKey).then(function success(response) {
     ctrl.post = response;
  });
}
```

for last but not least, the HttpService calls MessageService.showToast in the request's error function, so, we don't have to do that in the controllers anymore.

**TODO/FIXME:** Remove service.posts = response from postService.getNextPosts. Apparently this array isn't being used in anywhere.

### DIFF
--- a/frontend/post/postService.js
+++ b/frontend/post/postService.js
@@ -3,7 +3,7 @@
 (function() {
     var app = angular.module("app");
 
-    app.service("PostService", function PostService(HttpService, $q, AuthService) {
+    app.service("PostService", function PostService(HttpService, AuthService) {
         var service = this;
         var POSTS_URI = "/api/posts";
         var LIMIT = 10;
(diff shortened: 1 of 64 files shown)


frontend/post/postService.js
```diff
@@ -3,28 +3,21 @@
...
-            });
-            return deferred.promise;
+            return HttpService.get("/api/user/timeline");
         };
 
         service.getNextPosts = function getNextPosts(page) {
             var deferred = $q.defer();
-            $http.get("/api/user/timeline?page=" + page + "&limit=" + LIMIT).then(function success(response) {
-                service.posts = response.data;
+            HttpService.get("/api/user/timeline?page=" + page + "&limit=" + LIMIT).then(function success(response) {
```
Luiz-FS: The HttpService already returns a promise, you can return it.

raonismaneoto: For this situation the service sets its posts property in the success function, that's why it is resolving the promise and returning a new one...

PR #1167, opened 2018-06-14
Fix comment a post whose institution is not active

raonismaneoto:
**Feature/Bug description:**
It was allowed to comment in a post whose institution was inactive.
**Solution:**
I've added a verification in post_comment_handler to block this flow.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/utils/messageService.js
+++ b/frontend/utils/messageService.js
@@ -10,7 +10,6 @@
             "auth/email-already-in-use": "Email informado já está cadastrado.",
             "auth/wrong-password": "Senha incorreta ou usuário não possui senha.",
             "auth/user-not-found": "Usuário não existe.",
-            "Error! The institution has been deleted": "A instituição está inativa.",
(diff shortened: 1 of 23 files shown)


frontend/utils/messageService.js
```diff
@@ -38,7 +38,8 @@
             "Error! You've already voted in this survey": "Você já votou nessa enquete",
             "Error! User is not allowed to edit this post": "O usuário não tem permissão para editar essa publicação.",
             "Error! This post cannot be updated": "A publicação não pode ser editada.",
-            "Error! You don't have permission to publish post.": "Você não tem permissão para publicar post nesta instituição"
+            "Error! You don't have permission to publish post.": "Você não tem permissão para publicar post nesta instituição",
+            "Error! This institution is not active": "Essa instituição não está ativa"
```
mayzabeel: there is already a similar error on line 13

mayzabeel: use same message

PR #1185, opened 2018-06-18
Disabling button to send requests to be a member of an institution

tiagolimpe:
**Feature/Bug description:** When the user send a request to be a member of an institution the button to send request remains enabled to click.

**Solution:** Adding a simple list of keys of requested institutions by the user and verifying if the institution has been requested by the member.

screenshot from 2018-06-18 17-27-11
screenshot from 2018-06-18 17-27-33

**TODO/FIXME:** n/a

### DIFF
--- a/backend/handlers/user_handler.py
+++ b/backend/handlers/user_handler.py
@@ -33,7 +33,7 @@ def get_requests(user_key):
     """Query to return the list of requests that this user sent to institutions."""
     institutions_requested = []
     queryRequests = RequestUser.query(RequestUser.sender_key == user_key, RequestUser.status == 'sent')
-    institutions_requested = [request.institution_key.urlsafe() if request.institution_key.get().state == "active" else '' for request in queryRequests]
+    institutions_requested = [request.institution_key.urlsafe() for request in queryRequests]
 
     return institutions_requested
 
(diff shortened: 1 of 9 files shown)


frontend/requests/requestInvitationController.js
```diff
@@ -28,6 +28,8 @@
             var request = new Invite(dataInvite);
             var promise = RequestInvitationService.sendRequest(request, requestInvCtrl.institutionSelect.key);
             promise.then(function success() {
+                requestInvCtrl.currentUser.institutions_requested.push(requestInvCtrl.institutionSelect.key);
```
raonismaneoto: assert that it is been called in the function's test

backend/handlers/user_handler.py
```diff
@@ -29,6 +29,16 @@ def get_invites(user_email):
 
     return invites
 
+def get_requests(user_key):
+    """Query to return the list of requests that this user sent to institutions."""
+    institutions_requested = []
+
+    queryRequests = Invite.query(Invite.sender_key == user_key,
```
mayzabeel: is is good, check if this invite is a request to me member, because, the actual query get all invites sent by the user, can get invites for new institutions, links and invites for new members if the user is the admin

PR #1190, opened 2018-06-20
Create test event controller

MaianaB:
**Feature/Bug description:** The event controller didn't have a test.

**Solution:** Test created. The loadMoreEvent method needs to exist httpbackend with the route, but the return is indifferent, it is set to mock created for the service function.

**TODO/FIXME:** n/a


frontend/test/specs/event/eventControllerSpec.js
```diff
@@ -0,0 +1,170 @@
...
+                return {
+                    then: function (callback) {
+                        return callback(requestEventInst);
+                    }
+                };
+            });
+        });
+
+        it('should call getEvents', function(done) {
+            expect(eventCtrl.events.length).toBe(0);
```
Luiz-FS: * Once the controller is started the function loadMoreEvents is already called in main, so the list of events should not be empty.
* Use the toEqual () function to compare if the return is the same as expected.

frontend/test/specs/event/eventControllerSpec.js
```diff
@@ -0,0 +1,170 @@
...
+            var promise = eventCtrl.loadMoreEvents();
+            promise.should.be.fulfilled.then(function() {
+                expect(eventCtrl.events.length).toBe(2);
```
Luiz-FS: Check if this promise is actually being executed, I have already arranged tests where the same call (promise.should.be.fulfilled) existed and they were never executed and therefore always passed.

frontend/test/specs/event/eventControllerSpec.js
```diff
@@ -0,0 +1,170 @@
...
+        httpBackend = $httpBackend;
+        rootScope = $rootScope;
+        eventService = EventService;
+        messageService = MessageService;
+        mdDialog = $mdDialog;
+        state = $state;
+        AuthService.login(user);
+
+        $httpBackend.expect('GET', GET_EVENTS_URI).respond(requestEvent);
+        $httpBackend.expect('GET', GET_EVENTS_INST_URI).respond(requestEventInst);
```
Luiz-FS: Add the $httpBackend.flush() call to actually perform the http call.

frontend/test/specs/event/eventControllerSpec.js
```diff
@@ -0,0 +1,170 @@
...
+        eventCtrl = createCtrl();
+        eventCtrl.showImage = true;
+        eventCtrl.events = [];
+    }));
+
+    describe('main()', function() {
+
+        it('should call functions', function() {
+            eventCtrl = createCtrl();
+            expect(eventCtrl.posts.length).toBe(0);
```
Luiz-FS: Use toEqual().

frontend/test/specs/event/eventControllerSpec.js
```diff
@@ -0,0 +1,170 @@
...
+                };
+            });
+        });
+
+        it('should call getEvents', function(done) {
+            expect(eventCtrl.events.length).toBe(0);
+            var promise = eventCtrl.loadMoreEvents();
+            promise.should.be.fulfilled.then(function() {
+                expect(eventCtrl.events.length).toBe(2);
+                expect(eventCtrl.events[0].institution_key).toBe(institution.key);
```
Luiz-FS: Use toEqual()

frontend/test/specs/event/eventControllerSpec.js
```diff
@@ -0,0 +1,170 @@
...
+                expect(eventCtrl.events[1].institution_key).toBe(other_institution.key)
+            }).should.notify(done);
+            scope.$apply();
+
+            expect(eventService.getEvents).toHaveBeenCalledWith(0, undefined);
+        });
+
+        it('should call getInstEvent because the controller has institutionKey', function(done) {
+            eventCtrl.institutionKey = institution.key
+            expect(eventCtrl.events.length).toBe(0);
```
Luiz-FS: Same

frontend/test/specs/event/eventControllerSpec.js
```diff
@@ -0,0 +1,170 @@
...
+            eventCtrl.loadMoreEvents();
+            var promise = eventCtrl.loadMoreEvents();
+            promise.should.be.fulfilled.then(function() {
+                expect(eventCtrl.events.length).toBe(1);
+                expect(eventCtrl.events[0].institution_key).toBe(institution.key);
```
Luiz-FS: Some.

PR #1196, opened 2018-06-21
Refact function reload of auth service

MaianaB:
**Feature/Bug description:**  Refactoring the reload function used when is necessary to update user and save in cache. Before, this method created the new user. All notifications that need to be updated by the user must be described in the userService for the observer to be created, the ACCEPT_LINK notification being created observer in the AuthService.

**Solution:**  This method updates the properties of the user, so in notifications that need to refresh user call this method. Change notification to userService.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/auth/authService.js
+++ b/frontend/auth/authService.js
@@ -162,7 +162,7 @@
 
         service.updateUser = function updateUser(data){
             return Object.assign(userInfo, data);
-        }
+        };
 
         service.sendEmailVerification = function sendEmailVerification(user) {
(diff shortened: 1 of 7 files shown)


frontend/auth/authService.js
```diff
@@ -165,6 +160,10 @@
             return deferred.promise;
         };
 
+        service.updateUser = function updateUser(data){
+            return Object.assign(userInfo, data);
+        }
```
raonismaneoto: insert semicolon

PR #1203, opened 2018-06-28
Fix invalid request notification

ruanres:
**Feature/Bug description:** When the user clicks on a request notification that is related to an institution no more active, the notification dialog is shown with empty fields.

Exemple: User request
notificacao antiga

**Solution:** On RequestDialogService, in the selectDialog function, I added a verification to check if the request is valid, based on the request's institutions state. If on of the institutions is invalid, a dialog is showed to inform this to the user.

Invalid request dialog
invalid

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/institution/managementMembersController.js
+++ b/frontend/institution/managementMembersController.js
@@ -110,7 +110,7 @@
                 emails: emails
             };
             
-            if (manageMemberCtrl.isValidAllEmails(emails) && manageMemberCtrl.isUserInviteValid(invite)) {
+            if (manageMemberCtrl.isValidAllEmails(emails) && invite.isValid()) {
                 manageMemberCtrl.isLoadingInvite = true;
                 var promise = InviteService.sendInviteUser(requestBody);
                 promise.then(function success(responseData) {
@@ -124,6 +124,8 @@
                     manageMemberCtrl.isLoadingInvite = false;
                 });
                 return promise;
+            } else if(!invite.isValid()) {
+                MessageService.showToast('Convite inválido!');
             }
         };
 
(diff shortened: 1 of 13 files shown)


frontend/institution/managementMembersController.js
```diff
@@ -124,6 +124,8 @@
                     manageMemberCtrl.isLoadingInvite = false;
                 });
                 return promise;
+            } else if(!invite.isValid()) {
+                MessageService.showToast('Convite inválido!');
```
MaianaB: Não seria melhor colocar no message em português, e no controller em inglês ? Para manter o padrão.

frontend/requests/requestDialogService.js
```diff
@@ -82,18 +64,37 @@
...
+            let requestType = request.type_of_invite;
+            requestType = request.isStatusOn('sent') ? requestType : RESOLVED_REQUEST;
+            requestType = request.areInstitutionsValid() ? requestType : INVALID_REQUEST;
+
+            switch(requestType) {
+                case RESOLVED_REQUEST:
+                    service.showResolvedReqDialog(event); break;
+                case INVALID_REQUEST:
+                    service.showInvalidReqDialog(event); break;
+                case REQUEST_PARENT:
```
Luiz-FS: Why is it empty? should have at least one break.

ruanres: It is empty because both cases have the same behavior and it is not possible to do 
case REQUEST_PARENT || REQUEST_CHILDREN

frontend/institution/managementMembersController.js
ruanres: Seria, mas o MessageService mapeia as msg de erro do backend do inglês para português, e este não é o caso.

PR #1204, opened 2018-06-28
Send notification to refresh user when accept request

Luiz-FS:
**Feature/Bug description:** In certain system actions, such as administration transfer and request to create a link with other institutions, the permissions of the users involved need to be reloaded. Many problems are happening while reloading due to the fact that as soon as notification that a given request has been accepted, the user is reloaded and sometimes there is no time for the queue to add and remove permissions to finish, ie the user is reloaded but does not yet have the permissions. This is because the queue for adding permissions works asynchronously, just like the queue of notifications, so there is currently no way to control the synchronization between the various existing services.

**Solution:** Create a different queue type from the ones that already exist on the system and can be better managed to ensure synchronization of services.
I created a PULL queue where the activities are stored in it and then need to be removed and resolved, I also created a queue manager, it is responsible for creating a notification task, it returns a key that identifies the task newly created, and when it is necessary to solve the task simply pass your key to the manager that it will solve it. In this way, whenever it is necessary to create a notification that needs to be sent soon after some other queues task is resolved, this manager will be able to guarantee the synchronicity between the services.

**Resolved handlers:**
* Institution_parent_request_handler
* Institution_children_request_handler

**TODO/FIXME:** Remove 'if' in the addAdminPermissions queue when all handler notifications that pass through it are resolved.

### DIFF
--- a/backend/custom_exceptions/queueException.py
+++ b/backend/custom_exceptions/queueException.py
@@ -1,0 +1,10 @@
+"""Queue Exception."""
+
+__all__ = ['QueueException']
+
+class QueueException(Exception):
+    """Entity Exception."""
+
+    def __init__(self, msg=None):
+        """Init method."""
+        super(QueueException, self).__init__(msg or 'Invalid task')
--- a/backend/custom_exceptions/queueException.py
+++ b/backend/custom_exceptions/queueException.py
@@ -3,7 +3,7 @@
 __all__ = ['QueueException']
 
 class QueueException(Exception):
-    """Entity Exception."""
+    """Queue Exception."""
 
     def __init__(self, msg=None):
         """Init method."""
--- a/backend/handlers/institution_children_request_handler.py
+++ b/backend/handlers/institution_children_request_handler.py
@@ -49,10 +49,10 @@ class InstitutionChildrenRequestHandler(BaseHandler):
         institution_children = request.institution_requested_key.get()
         institution_children.set_parent(request.institution_key)
 
-        id_not = request.send_response_notification(user.current_institution, user.key, 'ACCEPT')
+        notification_id = request.send_response_notification(user.current_institution, user.key, 'ACCEPT')
         request.send_response_email('ACCEPT')
 
-        enqueue_task('add-admin-permissions', {'institution_key': institution_children.key.urlsafe(), 'id': id_not})
+        enqueue_task('add-admin-permissions', {'institution_key': institution_children.key.urlsafe(), 'id': notification_id})
 
         self.response.write(json.dumps(request.make()))
 
--- a/backend/handlers/institution_children_request_handler.py
+++ b/backend/handlers/institution_children_request_handler.py
@@ -49,7 +49,7 @@ class InstitutionChildrenRequestHandler(BaseHandler):
         institution_children = request.institution_requested_key.get()
         institution_children.set_parent(request.institution_key)
 
-        notification_id = request.send_response_notification(user.current_institution, user.key, 'ACCEPT')
+        notification_id = request.create_accept_response_notification(user.current_institution, user.key)
         request.send_response_email('ACCEPT')
 
         enqueue_task('add-admin-permissions', {
@@ -74,6 +74,6 @@ class InstitutionChildrenRequestHandler(BaseHandler):
         if institution_children.parent_institution == request.institution_key:
             institution_children.set_parent(None)
 
-        request.send_response_notification(user.current_institution, user.key, 'REJECT')
+        request.send_reject_response_notification(user.current_institution, user.key)
         request.send_response_email('REJECT')
         
--- a/backend/handlers/institution_children_request_handler.py
+++ b/backend/handlers/institution_children_request_handler.py
@@ -49,10 +49,12 @@ class InstitutionChildrenRequestHandler(BaseHandler):
         institution_children = request.institution_requested_key.get()
         institution_children.set_parent(request.institution_key)
 
-        request.send_response_notification(user.current_institution, user.key, 'ACCEPT')
+        id_not = request.send_response_notification(user.current_institution, user.key, 'ACCEPT')
         request.send_response_email('ACCEPT')
 
-        enqueue_task('add-admin-permissions', {'institution_key': institution_children.key.urlsafe()})
+        import pdb
+        pdb.set_trace()
+        enqueue_task('add-admin-permissions', {'institution_key': institution_children.key.urlsafe(), 'id': id_not})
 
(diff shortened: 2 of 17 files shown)


backend/util/notifications_queue_manager.py
```diff
@@ -0,0 +1,215 @@
...
+
+import json
+import time
+from google.appengine.api import taskqueue
+from service_messages import send_message_notification
+
+__all__ = ['NotificationsQueueManager', 'Notification', 'NotificationNIL']
+
+queue = taskqueue.Queue('notifications-manage-pull')
+notification_id = {
```
ruanres: I think that the ids would be more significant if they were the initials of its respective value

backend/util/notifications_queue_manager.py
```diff
@@ -0,0 +1,215 @@
...
+class NotificationsQueueManager:
+    """
+    Class to manage the notification queue.
+    """
+    
+    @staticmethod
+    def create_notification_task(notification):
+        """
+        Method for creating a task from the notification and 
+        trigger it in the queue. Returns the task key.
```
ruanres: ... and add it to the queue

backend/util/notifications_queue_manager.py
```diff
@@ -0,0 +1,215 @@
...
+        notification_type = notification_id[task_id[:2]]
+        task = find_task(notification_type,task_id)
+
+        if task:
+            notification = create_notification(task_id[:2], **json.loads(task.payload))
+            notification.send_notification()
+            queue.delete_tasks([task])
+
+
+class Notification(object):
```
ruanres: This class really need to be defined in this file?

backend/util/notifications_queue_manager.py
```diff
@@ -0,0 +1,215 @@
...
+        return self.__class__.__name__ + " " + str(formated_notification)
+    
+    def __repr__(self):
+        """
+        Method to generate string representation of notification object.
+        """
+        return str(self)
+
+
+class NotificationNIL(Notification):
```
ruanres: same question

backend/util/notifications_queue_manager.py
Luiz-FS: Placing initials can cause repeated ids.

backend/handlers/institution_children_request_handler.py
```diff
@@ -49,10 +49,13 @@ def put(self, user, request_key):
         institution_children = request.institution_requested_key.get()
         institution_children.set_parent(request.institution_key)
 
-        request.send_response_notification(user.current_institution, user.key, 'ACCEPT')
+        notification_id = request.send_response_notification(user.current_institution, user.key, 'ACCEPT')
```
mayzabeel: the send_response_notification method must be refactored, I think we should separate the creation of the notification and the sending in different methods when it is accepted the invite we call a method that creates and returns the id of the notification and when it is rejected we call send_response_notification

backend/models/request_institution_children.py
```diff
@@ -98,13 +100,24 @@ def send_response_notification(self, current_institution, invitee_key, action):
...
         )
 
-        super(RequestInstitutionChildren, self).send_notification(
-            current_institution=current_institution, 
-            sender_key=invitee_key, 
-            receiver_key=self.sender_key or self.admin_key,
-            notification_type=notification_type,
-            message=notification_message
-        )
+        if action == 'ACCEPT':
```
mayzabeel: separate creation and submission of notifications in different methods

backend/test/model_test/request_institution_children_test.py
```diff
@@ -213,13 +215,26 @@ def test_send_response_notification(self, create_notification_message, send_noti
...
             notification_type = 'ACCEPT_INSTITUTION_LINK' if action == 'ACCEPT' else 'REJECT_INSTITUTION_LINK'
-            send_notification.assert_called_with(
-                current_institution=self.other_institution.key, 
-                sender_key=self.other_institution.admin, 
-                receiver_key=self.admin.key, 
-                notification_type=notification_type,
-                message='mocked_message'
-            )
+
+            if action == 'ACCEPT':
```
mayzabeel: separate creation and submission of notifications in different methods

frontend/notification/notificationMessageCreatorService.js
```diff
@@ -44,8 +44,9 @@
             'REJECT_INVITE_USER_ADM': messageCreator('Rejeitou seu convite para ser administrador de ', SINGLE_INST),
             'ACCEPT_INVITE_HIERARCHY': messageCreator('Aceitou seu convite e estabeleceu vínculo entre ', DOUBLE_INST),
             'TRANSFER_ADM_PERMISSIONS': messageCreator('Você agora possui permissões para administrar ', SINGLE_INST),
+            'DELETED_USER': messageCreator('Removeu sua conta na Plataforma Virtual CIS', NO_INST),
+            'ADD_ADM_PERMISSIONS': messageCreator('Suas permissões hierarquicas foram atualizadas na instituição ', SINGLE_INST),
```
mayzabeel: hierárquicas

backend/worker.py
```diff
@@ -362,7 +363,13 @@ def addAdminPermissions(self, institution_key):
 
     def post(self):
         institution_key = self.request.get('institution_key')
+        id_nototification = self.request.get('id_notification')
```
mayzabeel: notification

backend/custom_exceptions/queueException.py
```diff
@@ -0,0 +1,10 @@
+"""Queue Exception."""
+
+__all__ = ['QueueException']
+
+class QueueException(Exception):
+    """Entity Exception."""
```
mayzabeel: Queue Exception

backend/models/request_institution_children.py
```diff
@@ -87,10 +89,28 @@ def send_notification(self, current_institution):
             notification_type=notification_type,
             message=notification_message
         )
+    
+    def create_accept_response_notification(self, current_institution, invitee_key):
+        """Send accept notification to sender of invite"""
```
mayzabeel: Create not send

backend/util/notification.py
```diff
@@ -0,0 +1,131 @@
...
+        return self.__class__.__name__ + " " + str(formated_notification)
+    
+    def __repr__(self):
+        """
+        Method to generate string representation of notification object.
+        """
+        return str(self)
+
+
+class NotificationNIL(Notification):
```
mayzabeel: for what?

Luiz-FS: For situations where the same worker may or may not send notifications on certain occasions.

PR #1212, opened 2018-07-03
Rename acceptInviteController and html to redirectInviteInstitutionController 

MaianaB:
**Feature/Bug description:** The accept Invite Controller had no role it accepted to invite, this controller redirected to an existing account or to create an account.

**Solution:** Changed name to redirect Invite Institution because this page is to institution invite and redirect according if user have account or not.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/test/specs/invites/acceptInviteControllerSpec.js
+++ b/frontend/test/specs/invites/acceptInviteControllerSpec.js
@@ -1,8 +1,8 @@
 'use strict';
 
-(describe('Test AcceptInviteController', function () {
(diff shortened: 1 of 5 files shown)


frontend/test/specs/invites/acceptInviteControllerSpec.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
-(describe('Test AcceptInviteController', function () {
+(describe('Test RedirectInviteInstitutionController', function () {
```
mayzabeel: change the name on this file

PR #1213, opened 2018-07-04
Send notification of accept transfer admin invite

Luiz-FS:
**Feature/Bug description:** When an administrator sends an administration transfer invitation to another member of the institution and he accepts, a notification must be sent to the current administrator and another to the new administrator as soon as the administration transfer queue finishes so that the permissions can be reloaded.

**Solution:** Use notifications queue manager to create a notification that will be resolved only when the administration transfer queue is finished.

**TODO/FIXME:** n/a

### DIFF
--- a/backend/models/invite_user_adm.py
+++ b/backend/models/invite_user_adm.py
@@ -6,6 +6,7 @@ from custom_exceptions import NotAuthorizedException
 from send_email_hierarchy import TransferAdminEmailSender
 from util import get_subject
 from util import Notification, NotificationsQueueManager
+from service_messages import create_system_message
 
 __all__ = ['InviteUserAdm']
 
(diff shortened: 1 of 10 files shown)


backend/models/invite_user_adm.py
```diff
@@ -63,9 +65,42 @@ def send_notification(self, current_institution):
             notification_type=notification_type
         )
     
-    def send_response_notification(self, current_institution, action):
+    def create_system_notification(self):
+        """."""
```
mayzabeel: describe this notification

backend/models/invite_user_adm.py
```diff
@@ -63,9 +65,47 @@ def send_notification(self, current_institution):
             notification_type=notification_type
         )
     
-    def send_response_notification(self, current_institution, action):
+    def create_system_notification(self):
+        """Create a new system notification to transfer admin permissions."""
```
mayzabeel: Create a new system notification for the new administrator to inform you that administrative permissions have been transferred.

backend/models/invite_user_adm.py
```diff
@@ -63,9 +65,47 @@ def send_notification(self, current_institution):
...
+            entity_key=self.key.urlsafe(),
+            notification_type="ACCEPT_INVITE_USER_ADM",
+            receiver_key=admin.urlsafe()
+        )
+
+        notification_id = NotificationsQueueManager.create_notification_task(notification)
+        return notification_id
+    
+    def send_reject_response_notification(self, current_institution):
         """Send notification to sender of invite when invite is accepted or rejected."""
```
mayzabeel: only is rejected no?

PR #1216, opened 2018-07-05
Fix worker aux functions test

ruanres:
**Feature/Bug description:** The single test of the suit of worker aux tests  was not being executed because its name did not begin with 'test_'

**Solution:** I renamed that test and added new ones for all the other aux functions. 

**TODO/FIXME:** n/a

### DIFF
--- a/backend/test/worker_handlers_tests/worker_aux_methods_test.py
+++ b/backend/test/worker_handlers_tests/worker_aux_methods_test.py
@@ -2,16 +2,171 @@
 """WorkerAuxMethodsTest."""
 
 from ..test_base import TestBase
-import permissions
 from .. import mocks
-from worker import get_all_parent_admins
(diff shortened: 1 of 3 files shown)


backend/test/worker_handlers_tests/worker_aux_methods_test.py
```diff
@@ -2,16 +2,98 @@
...
+        third_inst = mocks.create_institution()
+
+        user.add_institution(first_inst.key)
+        user.add_institution_admin(first_inst.key)
+
+        first_inst_urlsafe = first_inst.key.urlsafe()
+        second_inst_urlsafe = second_inst.key.urlsafe()
+        third_inst_urlsafe = third_inst.key.urlsafe()
+
+        # should be false, because user is admin of fist_inst
```
raonismaneoto: first*

PR #1224, opened 2018-07-10
Fix can't open institution request dialog

raonismaneoto:
**Feature/Bug description:**
It wasn't possible to open dialog when the notification's type was REQUEST_INSTITUTION. It was missing this type in the case inside the function that selects the dialog and the function areInstitutionsValid in invite.js was checking if the institutions were active instead of checking if the institutions weren't inactive.

screenshot from 2018-07-10 11-32-57

**Solution:**
I've added this case in the switch case and fixed the function.

screenshot from 2018-07-10 11-30-00

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/invites/invite.js
+++ b/frontend/invites/invite.js
@@ -28,9 +28,9 @@ Invite.prototype.setType = function (typeOfInvite) {
 };
 
 Invite.prototype.areInstitutionsValid = function () {
-    const isInstitutionActive = this.institution.isStateOn('active');
-    const isRequestedInstActive = this.requested_institution ? this.requested_institution.isStateOn('active') : true;
-    return isInstitutionActive && isRequestedInstActive;
+    const isNotInstitutionInactive = ! this.institution.isStateOn('inactive');
+    const isNotRequestedInstInactive = this.requested_institution ? ! this.requested_institution.isStateOn('inactive') : true;
+    return isNotInstitutionInactive && isNotRequestedInstInactive;
 }
 
 function isValueValid(value) {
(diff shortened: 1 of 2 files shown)


frontend/requests/requestDialogService.js
```diff
@@ -77,6 +77,8 @@
                 case REQUEST_PARENT:
                 case REQUEST_CHILDREN:
                     service.showHierarchyDialog(request, event); break;
+                case REQUEST_INSTITUTION:
```
ruanres: I do not understand why this is case necessary, once the default case has the same action and the REQUEST_INSTITUTION type should match with it.

frontend/invites/invite.js
```diff
@@ -28,9 +28,9 @@ Invite.prototype.setType = function (typeOfInvite) {
 };
 
 Invite.prototype.areInstitutionsValid = function () {
-    const isInstitutionActive = this.institution.isStateOn('active');
-    const isRequestedInstActive = this.requested_institution ? this.requested_institution.isStateOn('active') : true;
-    return isInstitutionActive && isRequestedInstActive;
+    const isNotInstitutionInactive = ! this.institution.isStateOn('inactive');
```
ruanres: I think this bug only occurs because of this function. It was not considering the case in which the state is 'pending', so the requestType was set to INVALID_REQUEST, at selectDialogToShow.

frontend/requests/requestDialogService.js
raonismaneoto: Because this approach is more readable and let it clearer that this case is treated in there.

PR #1237, opened 2018-09-14
Fix upload images and pdf

Luiz-FS:
**Feature/Bug description:** Due to firebase upgrade, uploading of images and PDFs was not working.

**Solution:** Update the way pictures and pdf are sent to the firebase.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/imageUpload/imageService.js
+++ b/frontend/imageUpload/imageService.js
@@ -69,10 +69,12 @@
             var promise = $firebaseStorage(image).$put(file, metadata);
 
             promise.$complete(function(snapshot) {
-                var data = {
-                    url: snapshot.downloadURL
-                };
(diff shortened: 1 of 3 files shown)


frontend/imageUpload/imageService.js
```diff
@@ -69,10 +69,12 @@
             var promise = $firebaseStorage(image).$put(file, metadata);
 
             promise.$complete(function(snapshot) {
-                var data = {
-                    url: snapshot.downloadURL
-                };
-                deferred.resolve(data);
+                snapshot.ref.getDownloadURL().then(function(downloadURL) {
```
raonismaneoto: can u test it to avoid involuntary removal later?

PR #1247, opened 2018-09-24
Push notifications setup

raonismaneoto:
**Feature/Bug description:**
Finish pushNotifications setup.

**Solution:**
In the client, the logic behind the user's permission token has been treated with firebase by saving the most recent token into the db. This only works to mobile devices.
In the server, the application retrieves the token by passing the user's ndb key.urlsafe() and using pyfcm send the notification to that token.
**TODO/FIXME:**
It's necessary to define where the notifications' properties are going to be created and when/what kind of notifications will be send.
Another issue is where to call the pushNotificationService. Injecting it in the notificationService, the way it is done now, it's not the best way to do that. But there aren't many options much less good options, because the service can only be started when the user login.
In the future we should intercept the notification in the service worker to personalize it. From there we can add vibration patterns, detailed actions and a lot of other options to increase the user's experience with push notifications.


backend/fcm.py
```diff
@@ -0,0 +1,40 @@
...
+def get_token(user_key):
+    response, content = _get_http().request(FIREBASE_TOKENS_ENDPOINT, method='GET')
+    data = json.loads(content)
+    if not user_key in data.keys():
+        return None
+    return token_filter(data, user_key)
+    
+def token_filter(data, user_key):
+    token_object = data[user_key]
+    token_object = token_object[token_object.keys()[0]]
```
JuliePessoa: always gonna be first object key? theres no word for the object you are looking for?
isn't better the above code?

>  `data[user_key]['first-key-i-dont-know-the-name']['token'] `

backend/requirements.txt
```diff
@@ -14,12 +14,13 @@ pyasn1==0.4.2
 pyasn1-modules==0.2.1
 PyJWT==1.5.2
 PyYAML==3.12
-requests==2.10.0
+requests==2.3
```
JuliePessoa: why downgrade?

frontend/notification/pushNotificationService.js
```diff
@@ -0,0 +1,90 @@
+(function() {
```
JuliePessoa: js doc

frontend/notification/pushNotificationService.js
```diff
@@ -0,0 +1,90 @@
...
+                    tokenObject.token = token;
+                    service.firebaseArrayNotifications.$add(tokenObject);
+                } else {
+                    currentTokenObject.token = token;
+                    service.firebaseArrayNotifications.$save(currentTokenObject);
+                }
+            });
+        }
+
+        function hasPermission() {
```
JuliePessoa: is it like Notification is enabled?

frontend/notification/pushNotificationService.js
```diff
@@ -0,0 +1,90 @@
...
+            const { permission } = Notification;
+            return permission === "granted";
+        }
+
+        (function init() {
```
JuliePessoa: how about test?

backend/fcm.py
```diff
@@ -0,0 +1,40 @@
+from pyfcm import FCMNotification
+
+from firebase_config import SERVER_KEY, FIREBASE_URL
+
+from firebase import _get_http
+
+import json
+
```
JuliePessoa: how about test?

frontend/index.html
```diff
@@ -75,7 +75,7 @@
   </script>
 
   <!-- AngularFire -->
-  <script src="https://cdn.firebase.com/libs/angularfire/2.3.0/angularfire.min.js"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/angularFire/2.3.0/angularfire.min.js"></script>
```
Luiz-FS: revert, this will conflict with the branch dev that has already been changed.

backend/fcm.py
```diff
@@ -0,0 +1,40 @@
...
+FIREBASE_TOKENS_ENDPOINT = "%s/pushNotifications.json" % FIREBASE_URL
+
+ICON_URL = "https://firebasestorage.googleapis.com/v0/b/eciis-splab.appspot.com/o/images%2FLOGO-E-CIS-1510941864112?alt=media&token=ca197614-ad60-408e-b21e-0ebe258c4a80"
```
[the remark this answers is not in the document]
raonismaneoto: Yes. This url is from production's firebase enviroment.

backend/fcm.py
```diff
@@ -0,0 +1,40 @@
...
+        return result
+
+def notify_multiple_users(tokens, title, body):
+    result = push_service.notify_multiple_devices(
+        registration_ids=tokens, message_title=title, message_body=body
+    )
+    return result
+
+def get_token(user_key):
+    response, content = _get_http().request(FIREBASE_TOKENS_ENDPOINT, method='GET')
```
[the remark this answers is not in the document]
raonismaneoto: _get_http() from firebase.py module is decorated by lru_cache that handles it.

backend/requirements.txt
raonismaneoto: fcm library wasn't compatible to that version, I had to change to the 2.3 version. I think none of our services has been affected by this change.

frontend/notification/pushNotificationService.js
raonismaneoto: yes, do you think the name is bad?

frontend/notification/pushNotificationService.js
raonismaneoto: I can test it, but for that, I'd need to hang some functions on the service to be able to call them in the test file, which is not a good approach because these functions would be available to be called by others, and we don't want it..

frontend/notification/pushNotificationService.js
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        /**
+         * It is responsible for check if the user already have a token.
+         * If it does, the token is replaced by the new one received as parameter.
+         * Otherwise the token is saved.
+         * @param {String} token 
+         * @param notificationsRef 
+         */
+        function setToken(token, notificationsRef) {
```
Luiz-FS: If the user enables notification on another device will the previous one be lost?

backend/test/util_modules_tests/fcm_test.py
```diff
@@ -0,0 +1,74 @@
...
+        self.assertTrue(tokens == [self.token])
+
+        self.firebase_object[self.s_user_key] = {
+            self.firebase_random_key: {
+                'token': 'test-token'
+            }
+        }
+
+        tokens = fcm.get_tokens([self.f_user_key, self.s_user_key])
+        self.assertEqual(tokens, [self.token, 'test-token'])
```
JuliePessoa: could be something like `self.token_f, self.token_s`

backend/test/util_modules_tests/fcm_test.py
```diff
@@ -0,0 +1,74 @@
...
+    
+    @patch('fcm.get_tokens_from_firebase')
+    def test_get_token(self, mocked_method):
```
JuliePessoa: shouldn't it be separated? like on test for None answer, another for existing answer. 
This is for all tests.

backend/worker.py
```diff
@@ -285,6 +287,11 @@ def post(self):
                     entity_key=post_key,
                     message=notification_message
                 )
+                
+                if is_first_like:
+                    title = "Primeira curtida"
```
JuliePessoa: It could be some kind of Enum. Notifications Types.

frontend/notification/pushNotificationService.js
JuliePessoa: We could use "_" to indicate that is a private function. along with @private notation.

frontend/notification/pushNotificationService.js
```diff
@@ -0,0 +1,108 @@
...
+            any: () => {
+                return (isMobile.Android() || isMobile.BlackBerry() || isMobile.iOS() || isMobile.Opera() || isMobile.Windows());
+            }
+        };
+
+        service.firebaseArrayNotifications;
+
+        service.currentUser = AuthService.getCurrentUser();
+
+        function requestPermission() {
```
JuliePessoa: js doc in all fucntions

frontend/notification/pushNotificationService.js
JuliePessoa: its something like, out of context it doesn't say what kind of permission you are verifying. something like `hasNotificationPermission` could say more. same thing with requestPermission

backend/fcm.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Args:
+        data: All the data retrieved from firebase.
+        user_key: The user ndb key.urlsafe().
+
+    Returns:
+        The filtered token.
+    """
+    token_object = data[user_key]
+    firebase_object = token_object.keys()[0]
```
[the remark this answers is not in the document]
raonismaneoto: If he changes the mobile, the token will update. Handle with 2 mobiles at the same time is the same problem we've discussed about cover others devices like chrome, mozilla... Besides, I think no one will use 2 mobile devices.

backend/fcm.py
```diff
@@ -0,0 +1,127 @@
...
+    data = get_tokens_from_firebase()
+    data_keys = data.keys()
+    tokens = []
+    for key in user_keys:
+        if key in data_keys:
+            current_token = token_filter(data, key)
+            tokens.append(current_token)
+    return tokens
+
+def get_tokens_from_firebase():
```
ruanres: you can get the user device tokens directly if you make the request with its urlsafe key. For example:
`request(.../pushNotifications/user-key.json)`

backend/worker.py
raonismaneoto: I'm improving it in another branch

backend/fcm.py
```diff
@@ -0,0 +1,165 @@
...
+        users_keys: The users' keys who will receive the
+        notification. 
+    
+    Returns:
+        The users' tokens.
+    """
+    tokens = []
+    for user_key in users_keys:
+        if user_key in content:
+            current_firebase_objects = content[user_key]
```
JuliePessoa: Is it possible to use the function above `filter_single_user_token` for each user?

frontend/notification/pushNotificationService.js
```diff
@@ -0,0 +1,123 @@
...
+                messaging.getToken().then(token => {
+                    service._saveToken(token);
+                });
+            });
+        };
+
+        /**
+         * It receives a token, starts a reference to
+         * firebase's database and save the token.
+         * @param {String} token
```
JuliePessoa: `@private` notation

frontend/notification/pushNotificationService.js
```diff
@@ -0,0 +1,123 @@
...
+            service._setToken(token, notificationsRef);
+        };
+
+        /**
+         * Instantiate a reference to the database
+         * based on the userKey, starts a firebaseArray
+         * in case of it hasn't been started before, and
+         * return the reference.
+         */
+        service._initFirebaseArray = function initFirebaseArray() {
```
JuliePessoa: same thing @private

frontend/notification/pushNotificationService.js
```diff
@@ -0,0 +1,123 @@
...
+        };
+
+        /**
+         * It is responsible for check if the user already have a token.
+         * If it does, the token is replaced by the new one received as parameter.
+         * Otherwise the token is saved.
+         * @param {String} token 
+         * @param notificationsRef 
+         */
+        service._setToken = function setToken(token, notificationsRef) {
```
JuliePessoa: same

frontend/test/specs/notification/pushNotificationServiceSpec.js
```diff
@@ -0,0 +1,113 @@
...
+            spyOn(service, '_saveToken').and.callFake(fakeCallback);
+
+            service._requestNotificationPermission();
+
+            expect(messaging.requestPermission).toHaveBeenCalled();
+            expect(messaging.getToken).toHaveBeenCalled();  
+            expect(service._saveToken).toHaveBeenCalled();
+        });
+
+        it('should not call saveToken', () => {
```
JuliePessoa: should not call, why?

frontend/test/specs/notification/pushNotificationServiceSpec.js
```diff
@@ -0,0 +1,113 @@
...
+        });
+
+        it('should call $add', () => {
+            spyOn(service.firebaseArrayNotifications, '$loaded').and.callFake(fakeCallback);
+            spyOn(service.firebaseArrayNotifications, '$add').and.callFake(fakeCallback);
+            
+            service._setToken(defaultToken, notificationsRef);
+
+            expect(service.firebaseArrayNotifications.$loaded).toHaveBeenCalled();
+            expect(service.firebaseArrayNotifications.$add).toHaveBeenCalled();
```
JuliePessoa: shouldn't it be good to test `haveBeenCalledWith`?

frontend/test/specs/notification/pushNotificationServiceSpec.js
```diff
@@ -0,0 +1,113 @@
...
+        });
+    });
+
+    describe('hasNotificationPermission', () => {
+        it('should return true', () => {
```
JuliePessoa: return true, why?
same thing for return false

frontend/test/specs/notification/pushNotificationServiceSpec.js
```diff
@@ -0,0 +1,113 @@
...
+            Notification = {permission: 'granted'};
+
+            const result = service._hasNotificationPermission();
+
+            expect(result).toBeTruthy();
```
JuliePessoa: is this case, should really be truthy? i think you actually want the `true` value. Same for false.

backend/fcm.py
```diff
@@ -0,0 +1,165 @@
...
+    Returns:
+        The users' token or an empty list when None 
+        of the users haven't enabled notifications yet.
+    """
+    data = get_all_tokens_from_firebase()
+    tokens = filter_multiple_user_tokens(data, users_keys)
+    return tokens
+
+
+def get_all_tokens_from_firebase():
```
JuliePessoa: what happens if any of these functions gets an error?

frontend/notification/pushNotificationService.js
```diff
@@ -0,0 +1,123 @@
...
+        /**
+         * Ask permission to the user to send push notifications
+         * and if the permission is conceded the user's new token
+         * is retrieved and saveToken is called passing the token
+         * as parameter.
+         */
+        service._requestNotificationPermission = function requestNotificationPermission() {
+            messaging.requestPermission().then(() => {
+                messaging.getToken().then(token => {
+                    service._saveToken(token);
```
JuliePessoa: what happens if it gets an error?

frontend/test/specs/notification/pushNotificationServiceSpec.js
raonismaneoto: In 'it' i'm simulating the case when the user doesn't enable the notifications.

frontend/test/specs/notification/pushNotificationServiceSpec.js
raonismaneoto: Yes, it would be good, but for that I need to access the tokenObject created inside the function. I can't create another one using $firebaseObject because each of them has a unique identifier

frontend/test/specs/notification/pushNotificationServiceSpec.js
raonismaneoto: Because the function tests if the user has given the permission, if so, the Notification.permission value is equal to granted.

frontend/test/specs/notification/pushNotificationServiceSpec.js
raonismaneoto: good point

frontend/notification/pushNotificationService.js
```diff
@@ -0,0 +1,130 @@
...
+
+            return notificationsRef;
+        };
+
+        /**
+         * It is responsible for check if the user already have a token.
+         * If it does, the token is replaced by the new one received as parameter.
+         * Otherwise the token is saved.
+         * @param {String} token 
+         * @param notificationsRef
```
JuliePessoa: type of notificationsRef?

frontend/test/specs/notification/pushNotificationServiceSpec.js
JuliePessoa: I'm sorry I wasn't clear. Its good to describe in the 'it' what's the test about.
So it should be something like `should not call saveToken when the user doesn't enable the notifications`

frontend/test/specs/notification/pushNotificationServiceSpec.js
JuliePessoa: same thing of the issue above.

backend/fcm.py
```diff
@@ -0,0 +1,164 @@
...
+    token property to the token's list.
+
+    Args:
+        content: A json returned from firebase.
+    
+    Returns:
+        The user's tokens.
+    """
+    tokens = []
+    for key in content:
```
Luiz-FS: Using map get better, for example:
```python
tokens = map(lambda key: key['token'], content)
```

backend/test/util_modules_tests/fcm_test.py
```diff
@@ -0,0 +1,107 @@
...
+                'token': 'token-2'
+            },
+            'opkdsfa-OFDO': {
+                'token': 'token-3'
+            }
+        }
+
+        tokens = fcm.filter_single_user_tokens(content)
+        self.assertTrue(len(tokens) == 3)
+        self.assertTrue(
```
Luiz-FS: Add assertion message

backend/test/util_modules_tests/fcm_test.py
```diff
@@ -0,0 +1,107 @@
...
+            'oakdopak-qopekqp': {
+                'token': 'token-2'
+            },
+            'opkdsfa-OFDO': {
+                'token': 'token-3'
+            }
+        }
+
+        tokens = fcm.filter_single_user_tokens(content)
+        self.assertTrue(len(tokens) == 3)
```
Luiz-FS: Add assertion message

backend/test/util_modules_tests/fcm_test.py
```diff
@@ -0,0 +1,107 @@
...
+                'oakdopak-qopekqp': {
+                    'token': 'token-2'
+                }
+            }
+        }
+
+        users_keys = [self.f_user_key, self.s_user_key, 'aop-OPAKSD']
+
+        tokens = fcm.filter_multiple_user_tokens(content, users_keys)
+        self.assertTrue(len(tokens) == 2)
```
Luiz-FS: Same

backend/test/util_modules_tests/fcm_test.py
```diff
@@ -0,0 +1,107 @@
...
+        self.assertTrue(
```
Luiz-FS: Same

frontend/notification/notificationService.js
```diff
@@ -4,7 +4,7 @@
     var app = angular.module("app");
 
     app.service("NotificationService", function NotificationService($firebaseArray, AuthService, $rootScope,
-        NotificationMessageCreatorService) {
+        NotificationMessageCreatorService, PushNotificationService) {
```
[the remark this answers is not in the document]
Luiz-FS: If calling the function ask for permission as soon as the user authenticates does not look better?

frontend/notification/pushNotificationService.js
```diff
@@ -0,0 +1,130 @@
...
+         * using Notification object.
+         * @private
+         */
+        service._hasNotificationPermission = function hasNotificationPermission() {
+            const { permission } = Notification;
+            return permission === "granted";
+        };
+
+        (function init() {
+            if (!service._hasNotificationPermission() && isMobile.any()) {
```
Luiz-FS: If the permission should only be requested as soon as the user logs it does not make sense to place the request in the init of the service.

backend/fcm.py
raonismaneoto: content is not a list, it's and object

PR #1253, opened 2018-09-27
Removed old version caches

JuliePessoa:
**Feature/Bug description:** 
Old version caches wasn't being removed from memory.

**Solution:** 
Made cache_sufix change with the app_version, so that the browser is able to detect file change in the sw.js. And added a "clearCache" function, that deletes old caches versions.

**TODO/FIXME:** n/a

### DIFF
--- a/ecis
+++ b/ecis
@@ -114,6 +114,17 @@ function set_frontend_url {
     set_config_file $line $service $url $service_name;
 }
 
+function set_app_version_config_sw {
+    sed -i "8s|.*|    APP_VERSION: '$1'|" $FRONTEND_CONFIG_FILE
+    catch_error $? "APP VERSION on Frontend $1"
+    sed -i "8s|.*|    APP_VERSION: '$1'|" $SUPPORT_CONFIG_FILE
+    catch_error $? "APP VERSION on Support $1"
+    sed -i "8s|.*|    APP_VERSION: '$1'|" $LANDINGPAGE_CONFIG_FILE
+    catch_error $? "APP VERSION on LandingPage $1"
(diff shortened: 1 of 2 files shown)


ecis
```diff
@@ -138,14 +160,22 @@ function catch_error {
...
 case "$1" in
+
+    commit)
+        git_commit
+    ;;
+
+    git-all)
+        git add .
+        git_commit
+        current_branch=$(git branch | grep "*" | tr -d "()*" | cut -c2-)
```
Luiz-FS: This command can be simplified using awk, for example, current_branch = $ (git branch | grep "*" | awk '{print $2}')

ecis
```diff
@@ -138,14 +160,22 @@ function catch_error {
...
+        git push origin $current_branch
+    ;;
+
     run)
         APP_VERSION=$(git branch | grep "*" | tr -d "()*" | cut -c2-)
```
Luiz-FS: Same

frontend/sw.js
```diff
@@ -27,6 +29,17 @@
         await myCache.addAll(urls);
     }
 
+    /**
+     * Clear old version caches.
+     */
+    function clearOldCache() {
+        caches.keys().then((keys) => {keys.filter((key) => !key.includes(CACHE_SUFIX)).map((key) => caches.delete(key))});
```
Luiz-FS: Return this promise

frontend/sw.js
```diff
@@ -27,6 +29,17 @@
...
+    }
+
+    self.addEventListener('activate', () => {
+        clearOldCache();
```
Luiz-FS: Receive event as a parameter and add event.waitUntil(clearOldCache()) so that the activation ends only when the cache is cleared.

PR #1255, opened 2018-10-01
Fix refresh user token

Luiz-FS:
**Feature/Bug description:** The way the user's login token was being reloaded is bad and does not work correctly. At each request, the token is reloaded, and if the user spends more than an hour without making a request, the token will expire and the user is de-logged.

**Solution:** Update the token in a certain time interval, before the token expires, instead of updating in every request.

**TODO/FIXME:** n/a


ecis
```diff
@@ -183,7 +183,7 @@ case "$1" in
         set_support_url local $SUPPORT_LOCAL
         set_frontend_url local $FRONTEND_LOCAL
 
-        dev_appserver.py $APP_YAML $FRONTEND_YAML $BACKEND_YAML $SUPPORT_YAML $WORKER_YAML $2
+        dev_appserver.py $APP_YAML $FRONTEND_YAML $BACKEND_YAML $SUPPORT_YAML $WORKER_YAML -A development-cis $2
```
raonismaneoto: It shouldn't be here, right?

frontend/auth/authService.js
```diff
@@ -8,21 +8,51 @@
...
+                });
+            } else if (tokenLoaded) {
+                return userInfo.accessToken;
+            }
+
+            return loadTokenPromise;
+        };
+
+        function getIdToken(user) {
+            user.getIdToken(true). then(function(userToken) {
```
raonismaneoto: there is a white space between then and the dot.

frontend/auth/authService.js
```diff
@@ -8,21 +8,51 @@
...
+                if (userInfo) {
+                    userInfo.accessToken = userToken;
+                    service.save();
+                    console.log(userToken);
```
raonismaneoto: remove

frontend/auth/authService.js
```diff
@@ -8,21 +8,51 @@
...
+                }
+
+                if (resolveTokenPromise)
+                    resolveTokenPromise(userToken);
+                    resolveTokenPromise = null;
+
+                tokenLoaded = true;
+            })
```
raonismaneoto: Is it missing a semicolon?

Luiz-FS: Fixed.

PR #1343, opened 2018-11-30
Fix refresh user token

Luiz-FS:
**Feature/Bug description:** The system was not working offline.

**Solution:** Correct the method of loading a new token to the user, so that if there is an error in the request, return the old token.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/test/specs/auth/authServiceSpec.js
+++ b/frontend/test/specs/auth/authServiceSpec.js
@@ -1,7 +1,7 @@
 'use strict';
 
 (describe('Test AuthService', function() {
-    var authService, userService, $q;
+    var authService, userService;
 
     var userTest = {
         name : 'User',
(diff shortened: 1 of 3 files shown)


frontend/test/specs/auth/authServiceSpec.js
```diff
@@ -84,4 +84,40 @@
...
+            const savedResolveTokenPromisse = authService.resolveTokenPromise;
+
+            authService._getIdToken(user).then(function(accessToken) {
+                expect(accessToken).toEqual(user.accessToken);
+                expect(savedResolveTokenPromisse).toHaveBeenCalled();
+                done();
+            });
+        });
+
+        it('should return actual token', function(done) {
```
JuliePessoa: describe why should return actual token

frontend/test/specs/auth/authServiceSpec.js
```diff
@@ -84,4 +84,40 @@
...
     });
+
+    describe('_getIdToken', function() {
+        beforeEach(function() {
+            authService.setupUser(userTest.accessToken);
+            authService.resolveTokenPromise = () => {};
+            spyOn(authService, 'resolveTokenPromise').and.callFake(() => {});
+        });
+
+        it('should refresh token', function(done) {
```
JuliePessoa: describe why should refresh token

PR #1346, opened 2018-12-03
Add ui router support in webchat

pedroespindula:
**Feature/Bug description:** Adds routing support to webchat with the following views:
- _**Home**_: When user is not calling neither chatting with anyone.
- _**Call**_: When user calls another user (audio only).
- _**Chat**_: When user chats with another user (chat only).
- _**Video**_: When user videocalls another user (video and audio).
- _**Login**_: When user is not logged in.

**Solution:** Use ui-router and $stateProvider defining each state and creating the correspondents empty views.

**TODO/FIXME:** Add auth support and views body.


webchat/app.js
```diff
@@ -2,8 +2,70 @@
...
+               url: "/",
+               views: {
+                   content: {
+                       templateUrl: "app/home/home.html",
+                       controller: "HomeController as controller",
+                   },
+               },
+           })
+           .state("webchat.chat", {
+               url: "/chat",
```
Luiz-FS: Extract all state name to constant.

PR #1365, opened 2018-12-18
Mark new institutions

raonismaneoto:
**Feature/Bug description:**
Insert a tag "novo cadastro" in institutions that the user logged in hasn't seen yet.

**Solution:**
Use localforage to control the new institutions.

screenshot from 2018-12-18 10-08-28

**TODO/FIXME:** n/a


frontend/institution/allInstitutionsController.js
```diff
@@ -89,11 +89,66 @@
...
+            });
             Utils.setScrollListener(content, allInstitutionsCtrl.loadMoreInstitutions);
         };
+
+        /**
+         * @private
+         * Get all institutions and insert them into localforage. This method
+         * is called when the cache is cleaned. After that, it calls loadMoreInstitutions.
+         */
+        allInstitutionsCtrl._setupSeenInstitutions = function _setupSeenInstitutions() {
```
JuliePessoa: I understand that, when you clen the cache, every institutions are being marked as new, thats right?
If it is, I think that its not necessary. better new not marked, than old marked.

frontend/institution/allInstitutionsController.js
```diff
@@ -89,11 +89,66 @@
...
+         * and check if each one has been seen by trying to retrieve
+         * it from the localforage. If the current institution is in the
+         * localforage, it has already been seen, otherwise, it hasn't and 
+         * the setItem is called what set the institution as seen in the next
+         * iteration.
+         * @param {array} institutions 
+         */
+        allInstitutionsCtrl._checkSeenInstitutions = function _checkSeenInstitutions(institutions) {
+            institutions.forEach(inst => {
+                return localforage.getItem(inst.key).then((successValue) => {
```
JuliePessoa: its better to save an array for the user, so you just have to do one get.

frontend/institution/allInstitutionsController.js
```diff
@@ -80,7 +80,7 @@
             .then(function success(response) {
                 currentPage += 1;
                 moreInstitutions = response.next;
-
+                allInstitutionsCtrl._checkSeenInstitutions(response.institutions);
```
JuliePessoa: shouldnt setInstitution depend on this?

frontend/institution/registered_institution.html
```diff
@@ -2,9 +2,16 @@
...
             style="background-image: url({{regInstCtrl.institution.cover_photo}});" id="card-cover-picture">
+            <div class="new-institution-tag-with-cover" md-colors="{background: 'teal-500'}" ng-if="regInstCtrl.hasNotSeenInstitution()">
+                <span><b>NOVO CADASTRO</b></span>
+            </div>
         </div>
-        <div ng-if="!regInstCtrl.hasCoverPhoto()" md-colors="{background: 'light-green'}">
+        <div ng-if="!regInstCtrl.hasCoverPhoto()" md-colors="{background: 'light-green'}" 
+            id="default-reg-inst-cover">
             <img ng-src="{{'app/images/institution.png'}}">
+            <div class="new-institution-tag" md-colors="{background: 'teal-500'}" ng-if="regInstCtrl.hasNotSeenInstitution()">
```
JuliePessoa: why its duplicated?

frontend/sw.js
```diff
@@ -6,6 +6,7 @@
     importScripts('https://storage.googleapis.com/workbox-cdn/releases/3.6.3/workbox-sw.js');
     importScripts('app/firebase-config.js');
     importScripts('app/config.js');
+    importScripts('https://cdnjs.cloudflare.com/ajax/libs/localforage/1.7.3/localforage.min.js');
```
JuliePessoa: is it being used?

backend/test/institution_handlers_tests/institution_collection_handler_test.py
```diff
@@ -39,11 +39,28 @@ def test_get(self, verify_token):
...
+
+    @patch('util.login_service.verify_token', return_value={'email': 'user@example.com'})
+    def test_get_all_institutions(self, verify_token):
+        """Test the get method."""
+        # create models
+        # new User
+        user = mocks.create_user('user@example.com')
+        
+        for i in xrange(15):
+            mocks.create_institution()
```
ruanres: I don't understand the point of using this for loop

frontend/institution/institutionService.js
```diff
@@ -9,8 +9,9 @@
         var LIMIT = 10;
         service.user = AuthService.getCurrentUser();
 
-        service.getInstitutions = function getInstitutions() {
-            return HttpService.get(INSTITUTIONS_URI);
+        service.getInstitutions = function getInstitutions(parameters) {
+            parameters = parameters ? parameters : '';
+            return HttpService.get(`${INSTITUTIONS_URI}${parameters}`);
```
ruanres: You can do
`${parameters || ''}`

frontend/localforageSetup.js
```diff
@@ -0,0 +1,12 @@
+(function () {
```
ruanres: I think it would be better to do
`(function startIDB() {
         ...
  })();`

frontend/test/specs/institution/allInstitutionsControllerSpec.js
```diff
@@ -128,4 +138,62 @@
             expect(institutionService.getNextInstitutions).toHaveBeenCalled();
         });
     });
+
+    describe('checkSeenInstitutions()', () => {
+        it('should call getItem and setItem', () => {
+            spyOn(localforage, 'setItem');
+            spyOn(localforage, 'getItem').and.callFake(() => {
```
ruanres: You can create a callFake function in the outer scope, so it is not necessary to define it for each test that is going to use it.

frontend/institution/allInstitutionsController.js
raonismaneoto: but there is no other way to do that, how would I know that an institution is new to put it in idb considering I don't have the old ones in another place?

frontend/institution/registered_institution.html
raonismaneoto: This tag, "novo cadastro", appears in the institution's cover, in this html there are two options for the cover. The first one is the institution's cover if there is one, otherwise the default cover is applied. That's the reason I need to duplicate the tag. Besides, the css is diferent once the default cover is a   and the first option is a background-image set in the style

backend/test/institution_handlers_tests/institution_collection_handler_test.py
raonismaneoto: I need more than 10 institutions created in the database to get them all. I've modified the handler to accept the option of get all institutions without pagination. The default limit value is 10

frontend/institution/allInstitutionsController.js
```diff
@@ -89,11 +89,66 @@
         /**
          * Start function.
          * Sets the scrollListener and retrieve the first 10
-         * institutions.
+         * institutions if the cache is not cleaned otherwise it calls
+         * setupInstitutions.
          */
         allInstitutionsCtrl.$onInit = function () {
-            allInstitutionsCtrl.loadMoreInstitutions();
+            localforage.length().then(size => {
```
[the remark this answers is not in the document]
raonismaneoto: yes it is. This other thing must use another instance as described in the documentation in the multiple instances section: https://localforage.github.io/localForage/

frontend/institution/allInstitutionsController.js
raonismaneoto: I didn't get it, what u mean by 'depend'?

PR #1389, opened 2019-01-17
Create institution links mobile page 

ruanres:
**Feature/Bug description:**
**Model**
model

**Solution:** 

**When there's no links**
no-links

**When there's only a patent institution**
parent

**When there's only children institutions**
sub

**When there's both**
both

**TODO/FIXME:** n/a


frontend/institution/institutional_links_mobile.html
```diff
@@ -0,0 +1,34 @@
+<div class="main-links-container">
+        
+    <div class="link-section" ng-if="instLinksCtrl.hasParentInst()">
```
JuliePessoa: these two divs looks pretty much the same, it could be a component and you just have to pass the array with the institutions

JuliePessoa: these two divs looks pretty much the same, it could be a component and you just have to pass the array with the institutions

frontend/app.js
```diff
@@ -189,7 +189,11 @@
                 url: "/institution/:institutionKey/institutional_links",
                 views: {
                     institution_content: {
-                        templateUrl: "app/institution/institutional_links.html",
+                        templateUrl: Utils.selectFieldBasedOnScreenSize(
+                            "app/institution/institutional_links.html", 
+                            "app/institution/institutional_links_mobile.html",
+                            420
```
JuliePessoa: use design screen constant

frontend/institution/institutionalLinks/instLinks/instLinks.component.js
```diff
@@ -0,0 +1,18 @@
+(function() {
+    "use strict";
+
+    angular
```
JuliePessoa: jsdoc

PR #1405, opened 2019-01-30
Temporary handler to create some missing attributes in production

raonismaneoto:
**Feature/Bug description:**
Some new attributes have been created what could lead to unexpected behavior at production enviroment.

**Solution:**
I've created a handler in admin.py that retrieve all users and institutions and add the missing attributes in the entities.

**TODO/FIXME:** n/a


raonismaneoto:
OBS: Só dar merge quando for necessário

PR #1406, opened 2019-01-30
Fix scrollbar on firefox to have ecis colors and thin width

pedroespindula:
**Feature/Bug description:** On firefox, the scrollbar is not defined as in other browsers, therefore, the custom ecis scrollbar does not render the way it should be.

**Solution:** Add firefox scrollbar definition.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/styles/custom.css
+++ b/frontend/styles/custom.css
@@ -602,6 +602,11 @@ a:active {
     background: #009688
 }
 
+.custom-scrollbar {
+    scrollbar-color: #009688 #ccc;
+    scrollbar-width: thin;
+}
+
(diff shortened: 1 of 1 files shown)


frontend/styles/custom.css
```diff
@@ -602,6 +602,11 @@ a:active {
     background: #009688
 }
 
+.custom-scrollbar {
+    scrollbar-color: #009688 #ccc;
```
JuliePessoa: it appears orange to me

JuliePessoa:
It will only work on firefox > 64

PR #1407, opened 2019-01-30
Webchat css refactor

pedroespindula:
**Feature/Bug description:** Disorganized and ineficient CSS full of visual bugs. Also, classes weren't according to BEM.

**Solution:** Add CSS grid, add CSS BEM classes declaration, split long CSS files in smaller ones.

**TODO/FIXME:** n/a


webchat/components/contacts/contacts-sidenav.css
```diff
@@ -0,0 +1,20 @@
+.contacts-sidenav {
+    height: 100%;
+    width: 100%;
+    display: grid;
+    grid-template-rows: 56px calc(100% - 56px);
```
brobsc: auto instead of calc seems to be enough to correctly space it.

pedroespindula: If calc is replaced for auto, the sidenav bugs on mobile

webchat/components/contacts/toolbar/contacts-toolbar.css
```diff
@@ -0,0 +1,11 @@
+.toolbar__container {
+    padding: 10px 0;
+    width: 100%;
+    display: flex;
```
raonismaneoto: couldn't u use grid?

PR #1430, opened 2019-02-11
Fix hide toolbars

MaianaB:
**Feature/Bug description:** The bottom toolbar 
**Solution:** The properties should reset only in mobile displays.

**TODO/FIXME:** n/a

### DIFF
--- a/frontend/utils/utils.js
+++ b/frontend/utils/utils.js
@@ -229,8 +229,10 @@ var Utils = {
      * In mode mobile maybe changes some properties.
      */
     resetToolbarDisplayStyle: function resetToolbarDisplayStyle(){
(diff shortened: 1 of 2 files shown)


frontend/utils/utils.js
```diff
@@ -229,8 +229,10 @@ var Utils = {
      * In mode mobile maybe changes some properties.
      */
     resetToolbarDisplayStyle: function resetToolbarDisplayStyle(){
-        document.getElementById('main-toolbar').style.display = 'block';
-        document.getElementById('navbar-bottom').style.display = 'flex';
+        if(Utils.isMobileScreen(450)){
```
JuliePessoa: Use the screen constant

PR #1475, opened 2019-02-20
Send email and notification for all institution admins

Luiz-FS:
**Feature/Bug description:** Send email and notification for all institution admins

**Solution:** Modify the backend so that emails and notifications are sent to all administrators of the institution.

**Completed flows:**

* send request institution parent
* send request instituion children

**TODO/FIXME:** n/a


backend/models/institution.py
```diff
@@ -59,6 +59,8 @@ class Institution(ndb.Model):
     # The admin user of this institution
     admin = ndb.KeyProperty(kind="User")
```
JuliePessoa, line 60: why the model still has to maintain the admin property?

JuliePessoa (changes requested):

backend/models/institution.py
raonismaneoto: @JuliePessoa he can't remove it now, if he do that, he will face many problems

raonismaneoto (approved):

backend/models/institution.py
Luiz-FS: So that the modifications can be divided into tasks without breaking the other features of the system, eventually all the features will be modified and the admin attribute will no longer be necessary and can be removed.

brobsc (approved):

JuliePessoa (approved):

PR #1496, opened 2019-02-28
Fixed empty poll option on mobile larger screens

### LINKED ISSUE #1471: Empty option is added on poll in larger mobile screen sizes.
**Bug Description**
Poll with only 2 options are being created with a third empty options on larger mobile screen sizes.
For example Iphone 6/7/8, Pixel 2

**To Reproduce**
1. On iphone 6/7/8 create a poll with only 2 options
2. See the poll on timeline with 3 options.

**Screenshots**
image

JuliePessoa:
**Feature/Bug description:** 
Fixes #1471 

On mobile larger screens like iphone 6/7/8 plus, polls with only two options fulfilled were posted with the third option empty.
It occurred because the code was removing an object array while was iterating in the same reference.
So what happened, the array contained 5 objects: [{"text"},{"text"},{""},{""},{""}].
Iterating over the array, every time it finds an empty object it was removed, but there were 3 empty equal objects and was the same referenced array.
On third iteration, the index = 2 , empty object removed. result array: [{"text"},{"text"},{""},{""}].
Next iteration, index = 3, so object 4 removed. result array: [{"text"},{"text"},{""}].
Theres no next iteration because theres no object on 4 index. That said, always would last one empty object.

**Bug screenshot**
image

**Solution:** 
Changed the way objects were removed using array.filter();

### DIFF
--- a/frontend/survey/survey.component.js
+++ b/frontend/survey/survey.component.js
@@ -49,8 +49,8 @@
         /* This method add ids in each option and remove the options that are empty.*/
         function modifyOptions(){
             surveyCtrl.options = surveyCtrl.options
-                .map((option, index) => {option.id = index; return option;})
-                .filter(option => option.text);
+                .filter(option => option.text)
+                .map((option, index) => {option.id = index; return option;});
         }
 
         function formateDate(){
--- a/frontend/survey/survey.component.js
+++ b/frontend/survey/survey.component.js
@@ -46,12 +46,15 @@
             return screen.width < 600;
         };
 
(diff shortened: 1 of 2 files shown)


JuliePessoa:
Fixes #1471

pedroespindula (approved):

raonismaneoto (approved):

raonismaneoto (changes requested):
Isnt there any tests for that function?

raonismaneoto (approved):

PR #1539, opened 2019-03-15
mobile dialog for pending invites 

ruanres:
**Feature/Bug description:** 
**Model**
dialog model

**Solution:** ...
dialogs
dialog-screens

**TODO/FIXME:** n/a


tiago-lp (approved):

JuliePessoa (changes requested):
try to make it fit better on larger screen. In this case an ascroll is not necessary
image

JuliePessoa (approved):

